PR #33, opened 2020-12-23
Supports "createMany" model method

### LINKED ISSUE #31: Support seeding the database
```js
import { factory, seed } from '@mswjs/data'

const db = factory(modles)

seed(db, {
  [modelName]: seedOptions
})
```

kettanaito:
## GitHub

- Closes #31 

## Changes

- [x] Adds `createMany` model method to create multiple random instances of the model.
- [ ] Supports multiple randomly created entities that reference the same relational model (#34).
- [ ] Type annotates a `oneOf` relational property when listing the `options.relations` in `createMany` (cannot create an entity without all its relational models specified, so `boolean` makes no sense).
- [ ] Supports creation of nested relational properties (user -> post -> category).

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -88,12 +88,9 @@ function createModelApi<
               if ('__type' in propDeclaration) {
                 const isManyOfRelationType =
                   propDeclaration.__type === RelationKind.ManyOf
-                const resolvedCount = isManyOfRelationType
-                  ? (count as number)
-                  : 1
                 const relationalModel = factory[propDeclaration.modelName]
                 const records = isManyOfRelationType
-                  ? relationalModel.createMany(resolvedCount)
+                  ? relationalModel.createMany(count as number)
                   : relationalModel.create()
 
                 acc[key] = records
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -6,6 +6,7 @@ import {
   EntityInstance,
   ModelDictionary,
   PrimaryKeyType,
+  RelationKind,
 } from './glossary'
 import { first } from './utils/first'
 import { executeQuery } from './query/executeQuery'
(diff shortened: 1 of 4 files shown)


kettanaito:
Hey, @tomaszgil. Thank you for your kinds words! I'm excited that MSW and its ecosystem helps you achieve your goals 🎉 

I'm not working on this feature currently. That being said, I still find this a crucial API to have, so if you're willing to help me out with this I'd be thankful. 

I _think_ some of the pre-requisites for `createMany` were implemented (like #34) but it may still need some discovery along the way. I haven't rebased this feature branch for some time, so it may make more sense to create a new one not to deal with the changes that have been introduced since it had been open (cherry-picking commits should work fine as we're adding a brand new feature here). 

I like that Prisma reference! This library is inspired by Prisma a lot, and modeling our `createMany` API around it would also be a good direction to follow:

```ts
const users = db.user.createMany([
  { name: 'Sonali', email: 'sonali@prisma.io' },
  { name: 'Alex', email: 'alex@prisma.io' },
])
```

- I don't think we need to nest initial values under the `data` property;
- Each item in the initial values array is identical to calling `db.user.create(initialValues)` individually. This includes the optional properties, relations, and using value getters from the model definition to generate the values that were not provided upon creation.

themagickoala:
Hi, I just found this after having noticed a gap in there not being a createMany method, and creating https://github.com/mswjs/data/pull/237 to address it. Feel free to close my PR if this is going to supercede it. Alternatively, if you like the route I've taken and would like me to adjust it to use an array format like Prisma, I'm happy to tweak it so this can get moved along.

themagickoala:
> Hi, I just found this after having noticed a gap in there not being a createMany method, and creating #237 to address it. Feel free to close my PR if this is going to supercede it. Alternatively, if you like the route I've taken and would like me to adjust it to use an array format like Prisma, I'm happy to tweak it so this can get moved along.

Have updated #237 to reflect the prisma-like syntax. Perhaps your initial approach with this branch could be satisfied in future by a `createList` model method akin to FactoryBot's syntax?

PR #45, opened 2021-01-07
Adds the "drop" database utility

### LINKED ISSUE #44: Clear all entities
I think that for testing purposes we need a way to delete all entities between tests. As an alternative, a user can loop all entities and call the delete method.

kettanaito:
- Closes #44


lavoscore:
I was facing issues trying to reset state in my tests, and I couldn't find a way to reset the db, and this `drop` saved my butt. But I had to venture deep into the woods to find this one. I really feel that this utility should be better documented in the readme.

Issue #49, opened 2021-02-05
Persistency layer


kettanaito:
I suggest we add a client-side synchronization layer to keep multiple clients that use the same DB in sync.

## Implementation

Each `factory` call attaches a sync middleware. That middleware establishes a `BroadcastChannel` with a reproducible ID that allows other DB to communicate via the same channel. 

Whenever there's a change to a DB (create/update/delete), that change is signaled to the channel. Other DB subscribe to channel events and apply the occurred change to their own instances. 

## Motivation

Multiple tabs of the same app should treat DB as a source of truth, meaning operations performed in one tab should be reflected when communicating with the DB in another tab. Since each JavaScript runtime will instantiate its own DB instance, the DB updates must be synchronized between clients. 

## Does this work in Node.js?

No. The synchronization layer is designed only for client-side usage. There is nothing to sync between the client and Node. That connection should be covered in the _persistance layer_. 

## Roadmap

- [x] Synchornize database operations between active tabs.
- [ ] Persist and hydrate database state in `sessionStorage`.

kettanaito:
This logic can be applied internally in two separate middleware:

- Synchronization middleware. Utilizes `BroadcastChannel` to sync database operations between currently open tabs.
- Persistency middleware. Flushes the latest database state into `sessionStorage` and hydrates from it upon page load. 

Those two are independent functionalities that together contribute to a good client-side experience.

easybird:
Hi, 

Thank you for this library. It's great. 
But this is the killer feature I need it to be usable..we want to use it during frontend development and some sort of quick demo mode.

@kettanaito , why did you stop working on #87  ?  Any plans on finishing it?
Or anything I can do to help finishing it?

kettanaito:
Hey, @easybird. Thank you for your kind words.

I think I've stopped working on the feature simply due to the lack of time. Technically, the last task I recall tackling was the proper serialization of the entire `database` instance to the session storage. Now, with the migration to Symbols for internal properties, it'd have to have an additional step to serialize those as well. 

I wouldn't advise following that particular branch, the library has changed quite significantly since then, including some internal refactoring that may pose more challenges than it's worth rebase upon. 

[...]

PR #66, opened 2021-04-14
Synchronizes database events between multiple clients

kettanaito:
## GitHub

- Relates to #49 

## Changes

- Adds a synchronization middleware that syncs database events (create/update/delete) between multiple open tabs.
- Each Database instance (internal) now has a unique reproducible MD5 hash to identify this instance among multiple database instances. This way a _single_ broadcast channel applies events to proper databases based on the database ID of the source event. 
  - MD5 hash is generated based on the following data:
    - Database invocation order. 
    - Database invocation location (a stack trace to where `factory()` is called. Factory creates an internal `Database` instance).

## Motivation

Client-side sync allows the database to feel like an actual external database you're working with, as its updates you provision in one tab are synced with the other tabs as if they are requesting the same database. In reality, each tab has its own runtime and creates its own database from a clean slate. 

## Roadmap

- [x] Ensure that multiple `db` instances sync their updates independently.


src/db/Database.ts
```diff
@@ -29,6 +34,21 @@ export class Database<Dictionary extends ModelDictionary> {
...
+
+  /**
+   * Generates a unique MD5 hash based on the database
+   * module location and invocation order. Used to reproducibly
+   * identify a database instance among sibling instances.
+   */
+  private generateId() {
+    const { stack } = new Error()
+    const callFrame = stack?.split('\n')[4]
+    const salt = `${callOrder}-${callFrame?.trim()}`
```
kettanaito, line 50: I'm using an invocation order and database location to identify a single database instance. 
The order of `factory()` calls is unlikely to change on runtime. Two first-created databases across clients aren't necessarily the same instances, as each client may load a different module, while the same `BroadcastChannel` controls _the entire host_. To account for that, I'm adding a module location to the hash salt. 

Two database instances on multiple pages are equal if:
1. They were called in the same order.
2. They were created in the same module.

kettanaito:
## Ignoring unrelated events

It's possible to have multiple `factory()` calls (and thus, `Database` instances) on a single page. The sync extension should understand that db has issued an update, and apply that update to the respective db in another client. This comes down to two possible solutions:

1. Create a unique `BroadcastChannel` per db instance.
2. Assign a unique reproducible ID to each db instance.

Unfortunately, `BroadcastChannel` is applied to the entire host and is discerned by channel name, not reference. Even if each db creates its own channel,

marcosvega91 (approved):
it looks very good :) nice 💯

Issue #56, opened 2021-04-07
Support relations when generating REST handlers


kettanaito:
## GitHub

- Follow up #12 

## What

Given the following model declarations:

```js
const db = factory({
  user: {
    id: primaryKey(String),
    posts: manyOf('post')
  },
  post: {
    slug: primaryKey(String)
  }
})
```

And called `db.user.toHandlers`, the following relation handlers must be created in addition to the default CRUD handlers:

- `GET /users/:id/posts`, returns all posts that belong to the user.
- `GET /users/:id/posts/:slug`, returns a post by a slug that belongs to the user.

I'd say that PUT and DELETE methods shouldn't be nested in this way, forcing you to update/delete a post instance directly (i.e. `PUT /posts/:slug`). 

## Questions

- Should `oneOf` relations pluralize the model name in the handler URL?

marcosvega91:
I think that for oneOf relations is good to have a singular noun. It makes the url self explanatory

kettanaito:
Makes perfect sense, @marcosvega91!

ruisaraiva19:
FYI: I'm working on this.

PR #67, opened 2021-04-14
feat: remove all internals from API model

### LINKED ISSUE #41: Omit internal properties when returning a public record
The properties described in the `InternalEntityProperties` interface must not be included in the publicly returned records:

https://github.com/mswjs/data/blob/3dc15fda9692aa897987cf17efd129d9b2645ec0/src/glossary.ts#L46-L49

People are likely to assert the publicly returned data directly, i.e. with `expect(db.user.findFirst(...)).toEqual({...})`. Including internal properties is implementation detail leakage and should be avoided.

marcosvega91:
In this PR I have removed all the internals from API.
To do so I have removed remove coupling of api with itself and add the map transformation on different method returns.

I have also removed the map from the rest handlers because it is already done in the api.

I have created a new `Entity` type, the public one, and replaced the old one with `InternalEntity`.

## GitHub

- Closes #41

## Changes

- Separates entities in `Entity` and `InternalEntity`.
- Adds an enum type for internal entity properties (`__type`, `__primaryKey`, etc.).
- Removes internal entity properties from public entities (i.e. from `db[model].create()` return payload).
- Sets `primaryKey` on the `Relation` when it's defined by looking it up in the database for the referenced model.
- Resolves relational properties against the `Relation` with no internal properties on the actual value (no longer needed, relation contains all the necessary info).
- Makes defined relational object properties _enumerable_ (#78).

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -1,14 +1,10 @@
 import md5 from 'md5'
 import { StrictEventEmitter } from 'strict-event-emitter'
-import {
-  InternalEntityInstance,
-  ModelDictionary,
-  PrimaryKeyType,
-} from '../glossary'
+import { InternalEntity, ModelDictionary, PrimaryKeyType } from '../glossary'
 
 type Models<Dictionary extends ModelDictionary> = Record<
   string,
-  Map<string, InternalEntityInstance<Dictionary, any>>
+  Map<string, InternalEntity<Dictionary, any>>
 >
 
 export type DatabaseMethodToEventFn<Method extends (...args: any[]) => any> = (
@@ -33,10 +29,7 @@ export class Database<Dictionary extends ModelDictionary> {
     this.events = new StrictEventEmitter()
     this.models = Object.keys(dictionary).reduce<Models<Dictionary>>(
       (acc, modelName) => {
-        acc[modelName] = new Map<
-          string,
-          InternalEntityInstance<Dictionary, string>
-        >()
+        acc[modelName] = new Map<string, InternalEntity<Dictionary, string>>()
         return acc
       },
       {},
@@ -64,7 +57,7 @@ export class Database<Dictionary extends ModelDictionary> {
 
   create<ModelName extends string>(
     modelName: ModelName,
-    entity: InternalEntityInstance<Dictionary, any>,
+    entity: InternalEntity<Dictionary, any>,
     customPrimaryKey?: PrimaryKeyType,
   ) {
     const primaryKey =
@@ -77,8 +70,8 @@ export class Database<Dictionary extends ModelDictionary> {
 
   update<ModelName extends string>(
     modelName: ModelName,
-    prevEntity: InternalEntityInstance<Dictionary, any>,
-    nextEntity: InternalEntityInstance<Dictionary, any>,
+    prevEntity: InternalEntity<Dictionary, any>,
+    nextEntity: InternalEntity<Dictionary, any>,
   ) {
     const prevPrimaryKey = prevEntity[prevEntity.__primaryKey]
     const nextPrimaryKey = nextEntity[prevEntity.__primaryKey]
@@ -112,7 +105,7 @@ export class Database<Dictionary extends ModelDictionary> {
 
   listEntities<ModelName extends string>(
     modelName: ModelName,
-  ): InternalEntityInstance<Dictionary, ModelName>[] {
+  ): InternalEntity<Dictionary, ModelName>[] {
     return Array.from(this.getModel(modelName).values())
   }
 }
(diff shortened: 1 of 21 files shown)


src/model/parseModelDeclaration.ts
```diff
@@ -67,12 +69,18 @@ export function parseModelDeclaration<
...
       const relationDefinition = declaration[key] as RelationDefinition<
-        RelationKind.OneOf,
+        any,
         ModelName
       >
-
-      if (exactValue) {
-        if (Array.isArray(exactValue)) {
+      if (exactValue && relationDefinition) {
+        const relationPrimarykey = findPrimaryKey(
```
marcosvega91, line 76: We can assume that primary key exists because we have already checked it when the factory is first created

kettanaito: That's a good point. I'd rather parse `primaryKey` in `parseModelDeclaration` and throw away `findPrimaryKey` function. It's confusing why the model that's known is parsed only when an entity is created. Needless to say, that creating new entities doesn't change model declaration's properties or relations. Looks like a design overhaul to me, and I'd like to address it in the follow-up pull request.

kettanaito:
Hey, @marcosvega91. Thank you for shipping this! I've rebased the feature branch to include the latest `which -> where` change. I will read through the changes, I'm curious how relations properties now understand that type of node value is being passed to them if we no longer expose its type publicly. A great chance for me to learn.

kettanaito:
This issue is more complicated than it appears. Since we've removed the internal properties from the entity, they are no longer present in the entity instance returned from the model API methods (create/update/etc.):

```js
const usa = db.country.create({ name: 'USA' })
```

This _is_ intended, as we wish for the public entity not to contain internal things. However, this means that relational properties cannot validate what kind of data you pass into them. Consider this:

```js
const db = factory({
  country: { id: primaryKey(String), name: String },
  capital: { id: primaryKey(String), country: oneOf('country') }
})

const usa = db.country.create({ name: 'USA' })

const notCountry = {
  id: 'abc-123',
  name: 'Made up'
}

// This is OK, as we pass an entity as the reltional value.
db.capital.create({ country: usa })

// Howver, this is also OK, because a non-entity object "notCountry"
// is compatible with its properties, although it's not the correct value to pass
// as it won't get resolved (doesn't have "__type", "__primaryKey", and "__nodeId"
// to look up the relational reference. 
db.capital.create({ country: notCountry })
```

I think we may design the internal logic in a different way, as all the necessary info for the relational lookup is known at the model declaration time:

```js
{
  country: {
    capital: manyOf('capital')
  }
}
```

Knowing that the referenced model is "capital" (`_type`), we can get its primary key. Then whenever you pass a value to the `country.capital`, we can expect that value to have a primary key and do a lookup only based on the value (type and primary key name known at the declaration time). 

@marcosvega91, what do you think about this? Was I successful in explaining the problem?

marcosvega91:
ah yes, I didn't think about it. When we'll create a new entity we should verify the existence of all relational entities

kettanaito:
@marcosvega91, yeah. I'll separate the concerns and make the relational properties work based on the model definiction, without us having to return the data like referenced entity `__type` and `__nodeId` (those are known when the property is defined). Will be addressed in #72.

marcosvega91:
Hey @kettanaito I think I have dropped one of your commit here. The one that you have pushed with the refactor of `removeInternalProperties` util. Could I ask you to cherry pick from your local ?

kettanaito:
@marcosvega91, no worries. I've applied that refactoring on top of your feature branch.

kettanaito:
Could you please tell me more about how you replace the entity references in relational properties?


Previously, the definition of the relational property produces a short `Relation` object:

```ts
factory({
  user: { country: oneOf('country') }
})

user.country // { kind: 'OneOf', modelName: 'country', unique: false }
```

Then, in` defineRelationalProperties`, when the actual entity values are known, they were used as an example of what that relation is:

https://github.com/mswjs/data/blob/39c078b569c213683f853364c071b5417ce8769d/src/model/defineRelationalProperties.ts#L31-L34

For instance,

```ts
db.user.create({
  country: db.country.create()
  // country: { __type: "country", __primaryKey: "id", ...initialValues }
})
```

With the removal of the internal entity properties like `__type` and `__primaryKey`, the latter won't be set on the actual value (won't be returned from `db[model].create`). I'd like to learn more about your thinking to tackle this. Thanks!

marcosvega91:
I have added the primary key in the [`Relation`](https://github.com/mswjs/data/blob/5cebdbc7f9af4b70ef6212c94d3426046e5343d3/src/glossary.ts#L41-L45) type. In this way when a new relation entity is created we already have all we need to reference it. The `__type` was already present as `modelName`

kettanaito:
Awesome! Got it now, thanks for the details.

kettanaito (approved):
A great improvement. Thank you, @marcosvega91!

marcosvega91:
This can be useful also for the issue #56, we'll have a better access to all the things we need to generate rest apis using relations

kettanaito:
True! I'm glad this touched relations a little, it will indeed give us more control over them in the future, as all the necessary data to resolve a relation is succinctly available on the `Relation` node.

PR #68, opened 2021-04-14
Rename which to where

### LINKED ISSUE #65: Consider renaming `which` to `where`
The property used to filter in mswjs/data is named `which`, citing [being heavily inspired by Prisma](https://twitter.com/ApiMocking/status/1381718391422132225). However, Prisma does not use `which`, it uses `where`:

https://www.prisma.io/docs/concepts/components/prisma-client/filtering-and-sorting

Is it possible that this is a mistake?

`where` is also the preferred word for other ORMs (presumably because the keyword is `WHERE` in SQL), such as:

- [Sequelize](https://sequelize.org/master/manual/model-querying-basics.html#applying-where-clauses)
- [TypeORM](https://typeorm.io/#/select-query-builder/what-is-querybuilder)
- [Firestore](https://firebase.google.com/docs/reference/js/firebase.firestore.Query#where)
- [Mirage](https://miragejs.com/api/classes/schema/#where)

Non-js:

- [SQLAlchemy (python)](https://docs.sqlalchemy.org/en/14/orm/queryguide.html#select-statements)
- [Rails' ActiveRecord](https://guides.rubyonrails.org/active_record_basics.html#read)
- [Ecto (elixir)](https://hexdocs.pm/ecto/Ecto.Query.html)

msutkowski:
Closes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ setupWorker(
   // Retrieve a single user from the database by ID.
   rest.get('/user/:userId', (req, res, ctx) => {
     const user = db.user.findFirst({
-      which: {
+      where: {
         id: {
           equals: req.params.userId,
         },
@@ -141,7 +141,7 @@ Returns the first entity that satisfies the given query.
 
 ```js
 const user = db.user.findFirst({
-  which: {
+  where: {
     id: {
       equals: 'abc-123',
     },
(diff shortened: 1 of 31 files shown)


kettanaito (approved):
Fantastic change. Thank you, Matt!
I'm one more sorry for overlooking your comment on this.

PR #70, opened 2021-04-15
REST handlers: Supports query parameters to query entity by its properties

### LINKED ISSUE #69: Support querying REST API resources by query parameters
## What

Support querying REST API resources by their properties through URL query parameters.

```js
const db = factory({
  car: {
    id: primaryKey(String),
   manufacturer: String
  }
})

setupServer(...db.car.toHandlers('rest'))

fetch('/cars?manufacturer=ford')
```

Where `?manufacturer=ford` is transformed to a `where` query against the `"manufacturer"` property of the car:

```js
modelApi.findMany({ where: { manufacturer: { equals: 'ford } } })
```

Internally, query parameters can be respected like so:

```js
// A very, very rough code.
rest.get(..., (req, res, ctx) => {
  const parameters = serialize(req.url.searchParams)
  const result = modelApi.findMany({ where: reduceParams(parameters) })
})
```

marcosvega91:
I have added filter parameters to the GET handler.

As described in the issue you can now do the following

`GET users?firstName=Kate&lastName=Moen`

close #69

### DIFF
--- a/src/model/generateRestHandlers.ts
+++ b/src/model/generateRestHandlers.ts
@@ -1,8 +1,15 @@
 import pluralize from 'pluralize'
 import { RestContext, RestRequest, ResponseResolver, rest } from 'msw'
-import { Entity, ModelDictionary, ModelAPI, PrimaryKeyType } from '../glossary'
-import { GetQueryFor } from '../query/queryTypes'
+import {
+  Entity,
+  ModelDictionary,
+  ModelAPI,
+  PrimaryKeyType,
+  ModelDefinition,
+} from '../glossary'
+import { GetQueryFor, QuerySelectorWhere } from '../query/queryTypes'
 import { OperationErrorType, OperationError } from '../errors/OperationError'
+import { findPrimaryKey } from '../utils/findPrimaryKey'
 
 interface WeakQuerySelectorWhere<KeyType extends PrimaryKeyType> {
   [key: string]: Partial<GetQueryFor<KeyType>>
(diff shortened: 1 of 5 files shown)


test/model/toRestHandlers.test.ts
```diff
@@ -126,10 +141,72 @@ describe('GET /users', () => {
...
+    expect(users).toEqual([
+      {
+        id: 'def-456',
+        firstName: 'Kate',
+        lastName: 'Moen',
+      },
+    ])
+  })
+
+  it('return all entities when wrong filter param is provided', async () => {
```
kettanaito, line 185: Shouldn't it return nothing, as in query didn't find any results?

marcosvega91: How I have thought to skip those parameters, checking if the param is in the model definition. 
We can return an error in this case ?

marcosvega91: I have looked around and seams that 400 bad request is the best way for this use case. What do you think ?

kettanaito: I think that's the right way to go!

I'd rephrase the error message to something like:

```
`Failed to query the "${modelName}" model: unknown property "${key}".`
```

Not sure about extending the `OperationErrorType` enum, though. It's meant to represent internal operation errors, while the wrong query parameter is a handler-specific case. I'd throw a regular error and mark it with a 400 status code:

```js
return res(ctx.status(400), ctx.json({ errorMessage: '...' }))
```

Do you think that would break the flow of how we handle the errors at the moment?

marcosvega91: To make error logic consistent I have extended the OperationError. In this way we can handle other errors in the `withErrors` funciton. Let me know what you think

kettanaito:
I'll try to rebase this against the lastest master and continue with the review.

src/model/generateRestHandlers.ts
```diff
@@ -72,12 +123,16 @@ export function generateRestHandlers<
         const cursor = req.url.searchParams.get('cursor')
         const rawSkip = req.url.searchParams.get('skip')
         const rawTake = req.url.searchParams.get('take')
```
kettanaito, line 125: Brainstorming here: do you think it would make sense to introduce a `parseQueryParams` function that would:

- Parse pagination properties and convert them to numbers, as needed.
- Return all the _rest_ properties and treat them as `filters`

```ts
function parseQueryParams(params: URLSearchParams): ParsedQueryParams

interface ParsedQueryParams {
  [property: string]: string
  take?: number
  skip?: number
  cursor?: PrimaryKeyType
}
```

It feels like we access query parameters multiple times but in an inconsistent manner.

marcosvega91: When working on filtes I was thinking the same, have a single entry point for it. I think that it makes sense

kettanaito (approved):
Thank you for adding this feature, @marcosvega91! 🎉

PR #75, opened 2021-04-19
Add generic type to getAll

camchenry:
Previously, getAll returned the type `any[]`, even though it returned additional properties. Now, the entity type is correctly inferred (similar to `findFirst`, `findMany`, etc.).

### DIFF
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -23,11 +23,11 @@ let callOrder = 0
 export class Database<Dictionary extends ModelDictionary> {
   public id: string
   public events: StrictEventEmitter<DatabaseEventsMap>
-  private models: Models<ModelDictionary>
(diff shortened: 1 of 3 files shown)


camchenry:
Looks I was a bit hasty to put up this PR. I was having some issues building/testing locally (fixed now), but it looks like there is a typing issue in the `Database` class.

kettanaito:
Hey, @camchenry. No worries. I've addressed the types of incompatibility and added a typings test case so we can catch regressions for this. I believe it's ready now.

kettanaito (approved):
This is a superb addition. Thank you, @camchenry, and welcome to contributors!

Let us know if you'd like to tackle some other open issues. We'd appreciate your feedback on them.

PR #77, opened 2021-04-20
chore: add @mswjs/data as a module

marcosvega91:
In tests `@mswjs/data` is imported from the lib folder.
It was giving a type error because the test folder was not include in tsconfig.

### DIFF
--- a/test/tsconfig.json
+++ b/test/tsconfig.json
@@ -1,10 +1,9 @@
 {
   "extends": "../tsconfig.json",
   "compilerOptions": {
-    "strict": true,
+    "declaration": false,
     "noEmit": true,
-    "esModuleInterop": true,
-    "downlevelIteration": true
+    "baseUrl": "../"
   },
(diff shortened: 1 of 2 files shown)


test/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "declaration": false,
```
kettanaito, line 4: Could you please include `"noEmit": true` so the tests don't get transpiled?

kettanaito (changes requested):
Thanks for fixing this, Marco! I've just left one comment about configuring TypeScript not to emit the test files, otherwise, this looks good.

test/tsconfig.json
marcosvega91: Yes sorry

kettanaito (approved):
Great job!

PR #79, opened 2021-04-21
fix: an entity with relations can be created also without them

marcosvega91:
If an entity has some relations as for model definition but it is created without them, the creation was throwing an error because it was trying to attach relations anyway.

I have created a new test file but the name is not really good I think 

## Changes

- It's now possible to create an entity without providing its relational properties (previously that would throw).


src/model/defineRelationalProperties.ts
```diff
@@ -36,6 +36,9 @@ export function defineRelationalProperties(
         `defining relational property "${entity.__type}.${property}"`,
         relation,
       )
+
+      if (!(property in initialValues)) return properties
```
kettanaito, line 40: This changes the established behavior, but I think it's a good change. Previously, it was an explicit decision to throw on missing relational properties, but looking at it now, I don't think that's necessary. You can attach the relational value later, which also makes those properties optional (just as any other entity properties).

In other words, where base values use value getters to populate created entities, relational properties must be explicitly set (they don't have a getter, instead they have a relational node produced by `oneOf`/`manyOf`). We should document this in README.

test/relations/empty.test.ts
```diff
@@ -0,0 +1,18 @@
+import { random } from 'faker'
+import { factory, primaryKey, oneOf } from '@mswjs/data'
+
+it('should not throw error if an entity with relation is created without it', () => {
```
kettanaito, line 4: I think this case should be added to all relevant `test/relations/*.test.ts` files, because the way oneOf and manyOf relations are handled may differ. Think of it as a spec point in each relationship kind support, not a standalone behavior.

src/model/defineRelationalProperties.ts
marcosvega91: do you think that something like this could be useful 

```diff
diff --git a/README.md b/README.md
index b9877fa..e04c19d 100644
--- a/README.md
+++ b/README.md
@@ -458,6 +458,8 @@ db.user.findFirst({
 
 Defining relationships is a way for one model to reference another. Models are flat by design, so if you wish to have a property that equals an object or an array, you create a relationship of the proper type to another existing model.
 
+All relations are optional as any other property specified in the model. You can create an entity and add relations later using the update API.
+
 #### One-to-One
 
 ```

Comment (kettanaito) [reply src/model/defineRelationalProperties.ts:40]: I think we need to elaborate on all the model properties being optional in the [create()](https://github.com/mswjs/data#create) section of the README. It also misses the example of calling `.create()` with initial values, which I think would be nice to showcase.

Comment (kettanaito) [reply src/model/defineRelationalProperties.ts:40]: I've pushed this addition to the README to this feature branch.

**Review by kettanaito — CHANGES_REQUESTED** [review]
This looks great, just need to structure the tests (add necessary test cases to the existing relationship tests).

**Comment by marcosvega91** [comment]
I had moved the test in their relation files but you made the same my mistake of the other day haahahaha, my last commit is dropped hahahah. I'll add it again, then I'll add some tests for missing properties.

**Comment by kettanaito** [comment]
@marcosvega91, I don't think I dropped anything... When I pushed to the feature branch its remote got 1 commit. This likely means that you committed something locally but haven't pushed yet. Meanwhile, I've pushed changes to the remote that you didn't have locally. When pulling, you probably gave priority to the remote changes, so your uncommitted local ones got discarded. 

What I usually do is `git pull --rebase` which tells me if I'm behind and ahead of the remote at the same time. This means I have new stuff but I also don't have some stuff from the remote. Then, depending on how many local commits I have, I follow a different rebase routine.

## If I have one local commit

```sh
$ git reset HEAD^
$ git stash -u
$ git pull --rebase
$ git stash pop
$ git commit -m 'The same message I had before'
```

## I have multiple local commits

```sh
$ git checkout -b temp
$ git checkout -
$ git pull --rebase --force
$ git rebase temp
```

marcosvega91:
You're as always, hahaha. I had updated the code but didn't commited yet. Now I have moved the test in the different relation files. I'll add some other test for other fields

kettanaito (approved):
Thank you for your work on this!

Issue #78, opened 2021-04-20
Error when calling findFirst even though nothing has changed


camchenry:
First, thanks for the great library. This solves the exact problem I have been working on, and we're already using MSW so this is a natural fit.

I have been running into an issue where the relations seem to "disappear" from the database, even though nothing has changed? I created a CodeSandbox which consistently crashes:
https://codesandbox.io/s/mswjs-relation-undefined-error-wkcfp?file=/src/App.tsx
```tsx
import { factory, oneOf, primaryKey } from "@mswjs/data";
import * as faker from "faker";

const db = factory({
  user: {
    id: primaryKey(faker.datatype.uuid),
    firstName: String
  },
  userObject: {
    id: primaryKey(faker.datatype.uuid),
    data: String,
    user: oneOf("user")
  }
});

// Seed the database
const seededUser = db.user.create({
  id: "ab4f631c-cca4-498f-a5aa-4828352a7c69",
  firstName: "Test"
});
db.userObject.create({
  user: seededUser,
  data: "test data 1 - associated with user"
});

const queryObject = () => {
  const object = db.userObject.findFirst({
    where: {
      user: {
        id: {
          equals: "ab4f631c-cca4-498f-a5aa-4828352a7c69"
        }
      }
    }
  });
  console.log("query:", { object, user: object.user });
};

// This will work OK, user is not undefined
queryObject();

// This will NOT work ok
setTimeout(() => {
  console.log("deferred query");
  queryObject();
}, 1000);
```

I make the same `queryObject` call twice: once immediately after creation, and another 1 second after seeding the database. My expectation is that both invocations of the function should do exactly the same thing. However, what happens is that the first call works as expected, while the second invocation crashes with the error `TypeError: can't access property "__type", actualValue is undefined`.

Willing to help out with this, let me know what questions you have. Thanks in advance.

(I apologize for all of the issues, but I am desperately trying to figure out this issue, and I have had no luck. I'd love to use this project.)

marcosvega91:
Hi @camchenry thanks for raising the issue :).

I think that making the property [`enumerable`](https://github.com/mswjs/data/blob/f8da857b3ea62dc2486b13f7cf9228e15e173b0d/src/model/defineRelationalProperties.ts#L82) should solve the issue. I have already done it in the same PR as for the other issue because I need it for deleting internal properties.

camchenry:
@marcosvega91 Sounds great! I will look forward to when that releases then.

kettanaito:
[...]

PR #81, opened 2021-04-21
Add async tests for findFirst and findMany

camchenry:
Added tests to ensure that both `findFirst` and `findMany` return identical results when called at a later point asynchronously, if no data has been mutated.

### DIFF
--- a/test/model/findFirst.test.ts
+++ b/test/model/findFirst.test.ts
@@ -1,5 +1,5 @@
-import { datatype } from 'faker'
-import { factory, primaryKey } from '@mswjs/data'
+import { datatype, seed } from 'faker'
+import { factory, oneOf, primaryKey } from '@mswjs/data'
 import { OperationErrorType } from '../../src/errors/OperationError'
 import { identity } from '../../src/utils/identity'
 import { getThrownError } from '../testUtils'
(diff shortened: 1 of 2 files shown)


camchenry:
@kettanaito @marcosvega91 This is a conversion of the code from the issue in #78. I added the test for both `findFirst` and `findMany` since those directly related to the original issue, but I'd like to know if there should be a similar test for any other APIs in your opinion.

test/model/findFirst.test.ts
```diff
@@ -90,3 +90,59 @@ test('returns null when found no matching entities', () => {
...
+    })
+  }
+
+  // Query for the object immediately
+  queryObject()
+
+  // This should return the exact same results as the query above,
+  // since nothing has changed
+  await new Promise<void>((res) =>
+    setTimeout(() => {
```
kettanaito, line 143: While this test showcases the usage example of when the issue was discovered, it doesn't reflect the intention behind the test. I think the intention is for the relational properties to be enumerable. There may be multiple use cases derived from that intention, just like you've added here.

Instead, we should add a test for `defineRelationalProperties.ts` that asserts that the relational properties defined on the entity object are enumerable (huh, we don't have any tests for that function!).

camchenry: @kettanaito Good point, I can look into writing some tests for `defineRelationalProperties` then. I suppose these tests are useful in that they validate the end-to-end behavior, but do you think we should just replace theses tests with `defineRelationalProperties` since that was the root cause?

kettanaito: Thank you! A single test that checks for the properties being enumerable is a good start, we can merge it at that state and iterate on it after. 

I think there's no need for the e2e test in this case, as there are no more systems involved around relational properties. If they are enumerable, you'll handle them as any other property. We may revisit this, but for now I'd say it's not necessary.

camchenry:
Going to close this PR and open a new one more focused on the relational properties.

kettanaito:
Thank you for carrying this out, @camchenry! This direction looks great—adding missing tests is always a pleasure.

PR #83, opened 2021-04-23
🏷️ add support for numeric primary keys

### LINKED ISSUE #82: `primaryKey` types does not support numbers
Hello! I'm trying out the package for the first time, and I've noticed that the primaryKey type does not support numbers:

https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/glossary.ts#L8

It is intentional? I need it to be number, as the real db my backend is using has numeric ids 😅 

I'll be happy to open a PR if needed ❤️

cloud-walker:
- Closes #82 

## Todos

- [x] run (and eventually fix) `yarn test:ts`
- [x] run (adn eventually fix) `yarn test`
- [x] run `yarn build`

### DIFF
--- a/src/model/generateRestHandlers.ts
+++ b/src/model/generateRestHandlers.ts
@@ -9,7 +9,7 @@ interface WeakQuerySelectorWhere<KeyType extends PrimaryKeyType> {
 }
 
 type RequestParams<Key extends PrimaryKeyType> = {
-  [K in PrimaryKeyType]: PrimaryKeyType
+  [K in Key]: string
 }
 
 export function createUrlBuilder(baseUrl?: string) {
--- a/src/utils/iteratorUtils.ts
+++ b/src/utils/iteratorUtils.ts
@@ -1,4 +1,6 @@
-export function forEach<K, V>(
+import { PrimaryKeyType } from "../glossary"
+
+export function forEach<K extends PrimaryKeyType, V>(
   fn: (key: K, value: V) => any,
   map: Map<K, V>,
 ): void {
@@ -7,7 +9,7 @@ export function forEach<K, V>(
   }
 }
 
-export function filter<K, V>(
+export function filter<K extends PrimaryKeyType, V>(
   predicate: (key: K, value: V) => boolean,
   map: Map<K, V>,
 ): Map<K, V> {
--- a/src/utils/iteratorUtils.ts
+++ b/src/utils/iteratorUtils.ts
@@ -1,4 +1,4 @@
-export function forEach<K extends string, V>(
+export function forEach<K, V>(
   fn: (key: K, value: V) => any,
   map: Map<K, V>,
 ): void {
@@ -7,7 +7,7 @@ export function forEach<K extends string, V>(
   }
 }
 
-export function filter<K extends string, V>(
+export function filter<K, V>(
   predicate: (key: K, value: V) => boolean,
   map: Map<K, V>,
 ): Map<K, V> {
--- a/src/db/Database.ts
+++ b/src/db/Database.ts
@@ -9,7 +9,7 @@ import {
 
 type Models<Dictionary extends ModelDictionary> = Record<
   string,
-  Map<string, InternalEntity<Dictionary, any>>
+  Map<PrimaryKeyType, InternalEntity<Dictionary, any>>
 >
 
 export type DatabaseMethodToEventFn<Method extends (...args: any[]) => any> = (
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -17,7 +17,7 @@ export enum InternalEntityProperty {
 
 export interface PrimaryKeyDeclaration {
   isPrimaryKey: boolean
-  getValue(): PrimaryKeyType
+  getValue(): KeyType
 }
 
 export enum RelationKind {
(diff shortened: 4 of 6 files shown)


cloud-walker:
I'm getting an error running `yarn test:ts`:

image

But I have no experience with @ts-expect-error directives

kettanaito:
The failing typings test is something odd. I think it's because building the library fails with these changes. Will elaborate below on how to fix the build.

https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/db/Database.ts#L12

Let's adjust ut to `Map<PrimaryKeyType, ...>` since entities are stored by their primary key, and it can now be number as well as string.

https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/query/executeQuery.ts#L35

Should accept `primaryKey: PrimaryKeyType` now.

https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/query/executeQuery.ts#L16

Also keeps entities as `Map<PrimaryKey, ...`.

https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/utils/iteratorUtils.ts#L1
https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/utils/iteratorUtils.ts#L10

Helper methods should also extend the primary key type in their generic: `<K extends PrimaryKeyType>`.

The last point concerns the `RequestParams` generic, which, I think, should also be `string` because request params are always strings, even when representing numbers:

https://github.com/mswjs/data/blob/8bb7419689dcbea52f5b758d4a302ab06e9310eb/src/model/generateRestHandlers.ts#L12

```ts
type RequestParams<Key extends PrimaryKeyType> = {
  [K in PrimaryKeyType]: string
}
```

I'm not entirely sure about the last one, if the querying will work properly if your model primary key is a number, then it gets to the handler as a string (params are strings), and then querying the database. @marcosvega91, do you have any concerns about this?

marcosvega91:
I have looked better at the type and I think there is something wrong in the use of `PrimaryKeyType`.

I have created a new PR on my fork with fixes that I think we should introduce to make it works
https://github.com/marcosvega91/data/pull/1

I think that the problem is that `PrimaryKeyType` should be used only for primary key value and not for key. The key should be always a string. `RequestParams` seams also wrong because key and value I think that are considered always string.

I hope to not say something wrong, let me know what you think <3

cloud-walker:
I don't know the difference between key and primaryKey in the codebase. My input was that there are database out there with numeric primary keys (like mysql for example).

Following this thought, I think the key should remain of type number as much as possible, and casted to string (for the `RequestParams` case for example, on the spot, only when necessary.

kettanaito:
@cloud-walker, your suggestions are correct. What I've mentioned above are some parts where instead of properly using `PrimaryKeyType` in the code, we had a hard-coded `string`. Your changes are a great chance for us to discover that, and fix it. Without that, the build won't pass.

src/utils/iteratorUtils.ts
```diff
@@ -1,4 +1,4 @@
-export function forEach<K extends string, V>(
```
cloud-walker, line 1: Here I've removed the extends part, instead of switching from `string` to  `PrimaryKeyType`: i think these utilities should remain generic, let me know if there is a better option 😉

src/model/generateRestHandlers.ts
```diff
@@ -9,7 +9,7 @@ interface WeakQuerySelectorWhere<KeyType extends PrimaryKeyType> {
 }
 
 type RequestParams<Key extends PrimaryKeyType> = {
-  [K in PrimaryKeyType]: PrimaryKeyType
```
cloud-walker, line 12: Here the `RequestParams` type was not using `Key`, I think it is an error

cloud-walker:
I'm currently getting a single error from the tests: 

image

I don't know how to solve it, yet.

(casting to string the user.id seems to work, but I guess is not the actual solution 😅 )

src/utils/iteratorUtils.ts
kettanaito: These utilities operate on the model collections, so the keys will always equal `PrimaryKeyType` anyway. I'd say extending that type guards us against improper usage of these utilities, and it must be preserved.

cloud-walker: Updated

cloud-walker:
I had the time to investigate the primaryKey.test problem, and it lies here: https://github.com/mswjs/data/blob/7248cb03bd388ce76e961e881c1a7f49d4e6a71d/src/query/queryTypes.ts#L50

As `user.id` is now a union the `GetQueryFor` type cannot distinguish the right query anymore, and interpret it as a string.

kettanaito:
Thanks for all the effort to make this better, @cloud-walker. I'll take a look at the failing test, it worked fine once I did those suggestions above, but I may have missed something. I will update that status here.

kettanaito:
So the issue was that the `primaryKey` function didn't inherit the exact value type but remained ambiguous `PrimaryKeyType`. I've added the value type inheritance to `primaryKey` and related types, so now it behaves correctly.

kettanaito (approved):
A great job done on this! Thank you for making numeric primary keys support happen, @cloud-walker! Welcome to contributors 🎉

PR #84, opened 2021-04-23
Add test to assert enumerable relational properties

camchenry:
Added a test to check that relational properties which get defined on the entity are enumerable. Checks #78 (fix added in #67).

### DIFF
--- a/test/model/relationalProperties.test.ts
+++ b/test/model/relationalProperties.test.ts
@@ -18,14 +18,14 @@ it('returns an enumerable relation property', () => {
   const db = new Database(dictionary)
 
   db.create('post', {
-    __primaryKey: 'id',
-    __type: 'post',
+    [InternalEntityProperty.primaryKey]: 'id',
+    [InternalEntityProperty.type]: 'post',
     id: 'post-1',
     title: 'First Post',
   })
   db.create('post', {
-    __primaryKey: 'id',
-    __type: 'post',
+    [InternalEntityProperty.primaryKey]: 'id',
+    [InternalEntityProperty.type]: 'post',
     id: 'post-2',
     title: 'Second Post',
   })
--- a/test/model/relationalProperties.test.ts
+++ b/test/model/relationalProperties.test.ts
@@ -1,4 +1,4 @@
-import { manyOf, primaryKey } from '../../src'
+import { manyOf, oneOf, primaryKey } from '../../src'
 import { Database } from '../../src/db/Database'
 import { InternalEntityProperty, Relation } from '../../src/glossary'
 import { defineRelationalProperties } from '../../src/model/defineRelationalProperties'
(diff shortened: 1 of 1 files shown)


camchenry:
@kettanaito Am I headed in the right direction with this PR? I'm still getting used to the internals of this library, so I welcome any feedback on how the test could be improved. I tried to distill it down to just the enumerable property part.

test/model/relationalProperties.test.ts
```diff
@@ -0,0 +1,52 @@
+import { manyOf, primaryKey } from '../../src'
+import { Database } from '../../src/db/Database'
+import { InternalEntityProperty, Relation } from '../../src/glossary'
+import { defineRelationalProperties } from '../../src/model/defineRelationalProperties'
+
+it('returns an enumerable relation property', () => {
```
kettanaito, line 6:
```suggestion
it('marks relational properties as enumerable', () => {
```

test/model/relationalProperties.test.ts
```diff
@@ -0,0 +1,52 @@
...
+  const user = {
+    [InternalEntityProperty.primaryKey]: 'id',
+    [InternalEntityProperty.type]: 'post',
+    id: 'abc-123',
+  }
+  const initialValues = {
+    posts: [{ id: 'post-1' }, { id: 'post-2' }],
+  }
+
+  defineRelationalProperties(user, initialValues, relations, db)
```
kettanaito, line 49: It's a shame this function requires so much boilerplate to be called. Nothing to tackle as a part of this pull request, but I'll think about how to make all these internal utils more testing-friendly.

test/model/relationalProperties.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  const db = new Database(dictionary)
+
+  db.create('post', {
+    [InternalEntityProperty.primaryKey]: 'id',
+    [InternalEntityProperty.type]: 'post',
+    id: 'post-1',
+    title: 'First Post',
+  })
+  db.create('post', {
```
kettanaito, line 26: Do you think it'd be a good idea to use `oneOf` relation just to reduce the number of entities we create in the test? It should result in the same behavior from the utility's perspective.

test/model/relationalProperties.test.ts
camchenry: Probably a better phrasing 👍

test/model/relationalProperties.test.ts
camchenry: Had the same thought while writing it. Took me a little while to actually figure out what "shape" of object needed to be passed into this function. Feels like there could be some improvements in the types there as well (maybe fewer usages of `any`)

test/model/relationalProperties.test.ts
camchenry: Yeah, that's a good point. Originally I actually had more tests but decided to pare it down, because it was sort of duplicating the other relational tests. Will rewrite to be a bit shorter perhaps.

kettanaito (approved):
This is a great addition! Thank you, @camchenry!

PR #90, opened 2021-04-28
feat: add snapshot utility

### LINKED ISSUE #86: Add a "snapshot" utility
## Expected behavior

The `snapshot` utility would memorize the state of the database (its entities) at a call time, and allow to restore to that memorized state at any further point of time (i.e. before each test).

## Usage

```ts
import { factory, primaryKey, snapshot } from '@mswjs/data'

const db = factory({
  user: {
    id: primaryKey(String),
    firstName: String
  }
})

db.user.create({ firstName: 'John' })
db.user.create({ firstName: 'Kate' })

// Takes a snapshot of the database
// in this point of time.
const restore = snapshot(db)

// Perform some other actions.
db.user.create({ firstName: 'Joe' })

// Restores the database to the taken snapshot.
// Will result in a database with 2 users:
// "John" and "Kate".
restore()
```

## GitHub

- Originated from #51 
- Related to #85

marcosvega91:
This is a first implementation of the snapshot utility. This will be updated after the implementation of the storage. 

close #86


kettanaito:
Hey, Marco! Thanks for putting this down.

I wouldn't merge this for now, as the issue isn't pressing and I will introduce some pre-requisite logic that we could then reuse in the `snapshot`. I'm currently working on the `Database` class serialization that's needed for the client-side persistence (#49), and I think the same serialization can be used for the `snapshot` utility. 

If you don't mind, let's put this on hold for the time being, and once the persistence lands, let's see what we can reuse.

micha149:
A bit late but I would like to suggest a slightly change to the api

```ts
import { factory, snapshot } from '@mswjs/data'

const db = factory({...})

// ...
// Set up some initial data
// ...

// Take a snapshot of all the entities in the DB and store it in a variable.
const applySnapshotWithInitialData = snapshot(db)

// Restores the DB to the previously taken snapshot. 
applySnapshotWithInitialData(db)
```
I think this will make a little more clear where the snapshot is applied to. Additionally it would enable us to create multiple instances and apply the snapshot to those as long as they have the same signature. One may call it copy-and-paste.

PR #98, opened 2021-06-07
feat: support primitive arrays in models

### LINKED ISSUE #97: Model doesn't support array of strings, numbers or booleans
Hi.

I cannot create models with array fields. The models support array of arrays or array of object, **but not array of strings, numbers or booleans**.

Here is an example:
```
import { primaryKey } from "@mswjs/data";

const car = {
  id: primaryKey((id: string) => id),
  name: (carName: string): string => carName,
  dimensions: () => [1, 2, 3], // <-- error
};
```

The browser gives the following error:
```
main.js:27438 Uncaught TypeError: Cannot use 'in' operator to search for '__type' in true
    at Object.isInternalEntity (isInternalEntity.js?0caa:9)
    at eval (removeInternalProperties.js?5ec5:44)
    at Array.map (<anonymous>)
    at eval (removeInternalProperties.js?5ec5:43)
    at Array.map (<anonymous>)
    at Object.removeInternalProperties (removeInternalProperties.js?5ec5:35)
    at Object.create (factory.js?898e:72)
    at handlers (person.mockRequest.ts?f8cc:34)
    at eval (handlers.ts?0e5c:21)
    at Array.flatMap (<anonymous>)
```

Am I doing something wrong or is this a bug?

Thanks!

with-heart:
Closes #97 

This PR adds primitive arrays support to models which is discussed in #97. 

Changes made to support this feature:
- created `BaseTypesArray` type
- added `BaseTypesArray` to `ModelDefinition` and `Limit`
- added check for primitive arrays in `createModel` to use the exact array value rather than executing `propertyDefinition()`
- added check for `object` entities to `manyOf` block in `removeInternalProperties` so that primitive array properties are left alone

This is my first time exploring `@mswjs/data`, so I could use some feedback on the approach here. 

I also ran into an issue where I was having to run `yarn build` in order for the tests to use the updated code. Is that typical?


src/utils/removeInternalProperties.ts
```diff
@@ -28,7 +28,10 @@ export function removeInternalProperties<
         }
 
         // Remove internal properties of a "manyOf" relation.
-        if (Array.isArray(value)) {
+        if (
+          Array.isArray(value) &&
+          value.some((entity: any) => typeof entity === 'object')
```
with-heart, line 33: Would it be better to use `!values.every(isPrimitive)` here? It's not entirely clear to me what all of the possible values in this array could be, so wasn't sure how to best check this.

test/model/create.test.ts
```diff
@@ -7,25 +7,40 @@ test('creates a new entity', () => {
   const db = factory({
     user: {
       id: primaryKey(identity(userId)),
+      name: () => 'Mark Chandler',
+      nicknames: () => ['with-heart'],
+      favoriteNumbers: () => [69, 420],
```
with-heart, line 12: I don't know that this is the best file to test this stuff in, but it's what made sense to me at the time. Can move/restructure as needed.

test/model/create.test.ts
```diff
@@ -7,25 +7,40 @@ test('creates a new entity', () => {
...
+  expect(randomUser).toHaveProperty('favoriteNumbers', [69, 420])
 })
 
 test('creates a new entity with initial values', () => {
   const db = factory({
     user: {
       id: primaryKey(datatype.uuid),
+      name: String,
+      nicknames: () => [] as string[],
+      favoriteNumbers: () => [] as number[],
```
with-heart, line 31: Is this fine? I'm not sure how users could define empty primitive arrays on models without specifying the type.

src/glossary.ts
```diff
@@ -73,7 +74,11 @@ export type ManyOf<ModelName extends KeyType> = RelationDefinition<
 
 export type ModelDefinition = Record<
   string,
-  (() => BaseTypes) | OneOf<any> | ManyOf<any> | PrimaryKeyDeclaration
+  | (() => BaseTypes)
+  | (() => BaseTypesArray)
```
kettanaito, line 78: Do you see any practical benefit of using a separate type than specifying the `BaseTypes[]` directly here?

```ts
(() => BaseTypes | BaseTypes[])
```

mo-klue:
@with-heart @kettanaito is there anything here that I can help with to make this PR ready to go?

src/utils/removeInternalProperties.ts
kettanaito: Actually, this check was dropped entirely in https://github.com/mswjs/data/commit/369a816b7eccf5025a7fb8356b18db268cd5f38e and we now allow anything that's not an internal entity, one or manyOf relation when removing internal properties. This should cover the arrays as well.

kettanaito: I've rebased your feature branch so this comment (and the change) is not applicable anymore.

test/model/create.test.ts
kettanaito: You've chosen the correct test suite: I think various data types support for the model declaration should be listed in the `create` test suite 👌

test/model/create.test.ts
kettanaito: That's a tricky question. Basically, the library infers whichever value type you give it. Since an empty array is likely result in `never[]`, I think the only way to denote a specific array item type without explicitly providing initial values is to use the type assertion (`as`) here.

kettanaito:
Currently the tests are failing here:

```sh
FAIL test/model/create.test.ts
  ● creates a new entity

    TypeError: Cannot use 'in' operator to search for '__type' in with-heart

       8 | function isInternalEntity(value) {
       9 |     return (value &&
    > 10 |         glossary_1.InternalEntityProperty.type in value &&
         |                                                ^
      11 |         glossary_1.InternalEntityProperty.primaryKey in value);
      12 | }
      13 | exports.isInternalEntity = isInternalEntity;

      at isInternalEntity (lib/utils/isInternalEntity.js:10:48)
          at Array.every (<anonymous>)
      at isManyOfRelation (lib/utils/removeInternalProperties.js:26:42)
      at lib/utils/removeInternalProperties.js:46:13
          at Array.reduce (<anonymous>)
      at Object.removeInternalProperties (lib/utils/removeInternalProperties.js:32:35)
      at Object.create (lib/factory.js:73:47)
      at Object.<anonymous> (test/model/create.test.ts:18:30)
```

This failure is caused by the `isInternalEntity` function being unable to process Arrays as entity values:

https://github.com/mswjs/data/blob/808295f6bb4a26f82a4085da9165da3bf4ac585c/src/utils/isInternalEntity.ts#L6-L14

So evaluating `"__type" in [1, 2, 3]` results in a type error (and rightfully so). I think we should extend `isInternalEntity` to understand arrays as entity values.

kettanaito:
I've introduced the object assertion in `isInternalEntity` as a separate pull request. Will rebase this feature branch once the said pull request lands.

kettanaito:
Array support in model definition was added in #113. Thank you for your work here, @with-heart!

PR #103, opened 2021-07-05
Supports sorting with the "orderBy" option on ".findMany()"

### LINKED ISSUE #102: Support sorting of records
First of all, thanks for this library! It really eases the process of mocking a database.

While implementing pagination in my mock REST API, I couldn't find any documentation about sorting. Does that feature exists and is not documented? If not, are there any plans to add it already or I'm the first one with a use case for it?

Thanks in advance!

kettanaito:
## GitHub

- Closes #102 

## Changes

It is now possible to sort query results of the `.findMany()` method using the `orderBy` property on the query:

```js
db.post.findMany({
  where: {
    category: {
      equals: 'Science',
    },
  },
  take: 10,
  orderBy: {
    title: 'asc',
  },
})
```

Provides support for:

- Basic sorting by direct properties.
- Sorting by multiple criteria.
- Sorting by relational properties. 

## Restrictions

- Only 1 known entity key can be specified in the sort criteria object. Use the list of criteria for multi-criteria sort. 

## Roadmap

- [x] Restrict multiple keys in the `orderBy` criteria object.
- [x] Improve the warning when using multiple keys in the `orderBy` criteria object (for non-TS users). 
- [x] Add typings tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ setupWorker(
 - [Strict mode](#strict-mode)
 - [Model relationships](#model-relationships)
 - [Pagination](#pagination)
+- [Sorting](#sorting)
 
 ### Model methods
 
@@ -640,6 +641,100 @@ const secondPage = db.post.findMany({
   // The second page will start from the last post
   // of the `firstPage`.
   cursor: firstPage[firstPage.length - 1].id,
+})
+```
+
+### Sorting
+
+#### Basic sorting
+
+```js
+const db = factory({
+  post: {
+    id: primaryKey(String),
+    title: String,
+  },
+})
+
+// Return first 10 posts in the "Science" category
+// sorted by the post's "title".
+db.post.findMany({
+  where: {
+    category: {
+      equals: 'Science',
+    },
+  },
+  take: 10,
+  orderBy: {
+    title: 'asc',
+  },
+})
+```
+
+> You can use `orderBy` with [pagination](#pagination).
+
+#### Sorting by relational properties
+
+```js
+const db = factory({
+  post: {
+    id: primaryKey(String),
+    title: String,
+    author: oneOf('user')
+  },
+  user: {
+    id: primaryKey(String),
+    firstName: String
+  }
+})
+
+// Return all posts in the "Science" category
+// sorted by the post author's first name.
+db.post.findMany({
+  where: {
+    category: {
+      equals: 'Science'
+    }
+  }
+  orderBy: {
+    author: {
+      firstName: 'asc'
+    }
+  }
+})
+```
+
+#### Sorting by multiple criteria
+
+Provide a list of criteria to sort the query result against.
+
+```js
+db.post.findMany({
+  orderBy: [
+    {
+      title: 'asc',
+    },
+    {
+      views: 'desc',
+    },
+  ],
+})
+```
+
+You can also use a combination of direct and relational properties on a single query:
+
+```js
+db.post.findMany({
+  orderBy: [
+    {
+      title: 'asc',
+    },
+    {
+      author: {
+        firstName: 'asc',
+      },
+    },
+  ],
 })
 ```
 
--- a/src/query/sortResults.ts
+++ b/src/query/sortResults.ts
@@ -47,6 +47,9 @@ function flattenSortCriteria<EntityType extends Entity<any, any>>(
   }, [])
 }
 
+/**
+ * Sorts the given list of entities by a certain criteria.
+ */
 export function sortResults<EntityType extends Entity<any, any>>(
   orderBy: OrderBy<EntityType> | OrderBy<EntityType>[],
   data: InternalEntity<any, any>[],
--- a/src/query/sortResults.ts
+++ b/src/query/sortResults.ts
@@ -1,0 +1,89 @@
+import debug from 'debug'
+import get from 'lodash.get'
+import { Entity, InternalEntity } from 'src/glossary'
+import { OrderBy, SortDirection } from './queryTypes'
+
+const log = debug('sortResults')
+
+type FlatSortCriteria = [string[], SortDirection]
+
+function warnOnIneffectiveSortingKeys(sortCriteria: Record<string, any>): void {
+  const [mainCriteria, ...siblings] = Object.keys(sortCriteria)
+
+  if (siblings.length > 0) {
+    console.warn(
+      'Sorting by "%s" has no effect: already sorted by "%s".',
+      siblings.join(','),
+      mainCriteria,
+    )
+  }
+}
+
+function flattenSortCriteria<EntityType extends Entity<any, any>>(
+  orderBy: OrderBy<EntityType>[],
+  propertyPath: string[] = [],
+): FlatSortCriteria[] {
+  log('flattenSortCriteria:', orderBy, propertyPath)
+
+  return orderBy.reduce<FlatSortCriteria[]>((criteria, properties) => {
+    warnOnIneffectiveSortingKeys(properties)
+
+    // Multiple properties in a single criteria object are forbidden.
+    // Use the list of criteria objects for multi-criteria sort.
+    const property = Object.keys(properties)[0] as keyof OrderBy<EntityType>
+    const sortDirection = properties[property]!
+    const path = propertyPath.concat(property.toString())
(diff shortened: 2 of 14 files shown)


kettanaito:
[Useful suggestion](https://stackoverflow.com/questions/57571664/typescript-type-for-an-object-with-only-one-key-no-union-type-allowed-as-a-key) for implementing a `OneKey` type that restricts an object to have only one of its known keys. Doesn't work with the complex nature of the `OrderBy` that represents objects with _relational properties_, which should also have only one known key.

kettanaito:
I've used this `RequireExactlyOne` type from `type-fest` to achieve only one property restriction on the sort criteria object:

https://github.com/sindresorhus/type-fest/blob/825bcaf7e0d1dedc7f453141f634b4fac36d0b14/source/require-exactly-one.d.ts

README.md
```diff
@@ -643,6 +644,100 @@ const secondPage = db.post.findMany({
...
+// Return all posts in the "Science" category
+// sorted by the post author's first name.
+db.post.findMany({
+  where: {
+    category: {
+      equals: 'Science'
+    }
+  }
+  orderBy: {
+    autor: {
```
fildon, line 700: Typo here on 'autor'?

```suggestion
    author: {
```

kettanaito: Thank you for spotting! Fixed.

src/query/sortResults.ts
```diff
@@ -0,0 +1,91 @@
+import debug from 'debug'
+import get from 'lodash.get'
+import { InternalEntity } from 'src/glossary'
+import { OrderBy, SortDirection } from './queryTypes'
+
+const log = debug('sortResults')
+
+type FlatSortCriteria = [string[], SortDirection]
+
+export function warnOnIneffectiveSortingKeys(obj: Record<string, any>): void {
```
fildon, line 10: I see a lot of 'any' in this codebase, but generally 'unknown' is safer
```suggestion
export function warnOnIneffectiveSortingKeys(obj: Record<string, unknown>): void {
```
There are legitimate uses for 'any' but as a rule of thumb you should start with 'unknown' until you are sure you need 'any'.

fildon, line 10: This function is only referenced in this file. No need to export it:
```suggestion
function warnOnIneffectiveSortingKeys(obj: Record<string, any>): void {
```

src/query/sortResults.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const criteriaList = ([] as OrderBy<any>[]).concat(orderBy)
+  log('criteria list:', criteriaList)
+
+  const criteria = flattenSortCriteria(criteriaList)
+  log('flattened criteria:', JSON.stringify(criteria))
+
+  data.sort((left, right) => {
+    for (const [path, sortDirection] of criteria) {
+      log('sorting "%s" ("%s")', path, sortDirection)
```
Aprillion, line 65: this log is probably not needed given the log on line 70 (unless lodash.get has a really terrible error message if it explodes)

kettanaito:
Typings tests fail due to TypeScript versions mismatch. Have it passing in the IDE, fails both on local CLI and CI. 

- VS Code: typescript@4.2.4
- CLI: typescript@4.3.5

src/query/sortResults.ts
```diff
@@ -0,0 +1,91 @@
...
+    console.warn(
+      'Sorting by "%s" has no effect: already sorted by "%s".',
+      siblings.join(','),
+      mainCriteria,
+    )
+  }
+}
+
+function flattenSortCriteria(
+  orderBy: OrderBy<any>[],
```
fildon, line 23: This 'any' is hiding some sneaky type stuff going on.

Really this function ought to be generically typed and then we can check against that generic type rather than need to use any:

```ts
function flattenSortCriteria<Entity>(
  orderBy: OrderBy<Entity>[],
```

Then on lines 33-35:
```ts
    const property = Object.keys(properties)[0] as keyof OrderBy<Entity>
    const sortDirection = properties[property]!
    const path = propertyPath.concat(property.toString())
```

src/query/sortResults.ts
Aprillion: also, printing `n * log(n) * numberOfCriteria` messages can be a bit too much logging even for `debug` level 😬

sorting tens of thousands of records without logs should be still fast,, but trying to log it might make it super slow - do you expect `debug` level to be used only by library maintainers (=> ok to debug in the loop) or is it a feature for users (=> might be better to hide the logging behind an extra flag)?

Aprillion (approved):

src/query/sortResults.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // Recursively flatten order criteria when referencing
+    // relational properties.
+    const newCriteria =
+      typeof sortDirection === 'object'
+        ? flattenSortCriteria([sortDirection], path)
+        : [[path, sortDirection]]
+
+    log('pushing new criteria:', newCriteria)
+    return criteria.concat(newCriteria as any)
```
fildon, line 46: This 'as' is pretty gnarly

If you change line 43 to:
```ts
: [[path, sortDirection]] as FlatSortCriteria[]
```

Then you can remove this `as any` here

src/query/sortResults.ts
```diff
@@ -0,0 +1,91 @@
...
+  }, [])
+}
+
+export function sortResults(
```
fildon, line 50: We can avoid 'any' in this function too:

```ts
export function sortResults<Entity>(
  orderBy: OrderBy<Entity> | OrderBy<Entity>[],
  data: InternalEntity<Record<string, unknown>, string>[],
): void {
  log('sorting data:', data)
  log('order by:', orderBy)

  const criteriaList = ([] as OrderBy<Entity>[]).concat(orderBy)
```

fildon (changes requested):

src/query/sortResults.ts
kettanaito: That's a valid concern. I believe `any` was adopted because `unknown` throws type violations in a strict (non-strict?) mode for some users.

Aprillion:
In the [force-pushed](https://github.com/mswjs/data/compare/2b0539ac7ca23071bc9984d215686c8b6c7385a9..7dbb3b5282c9b6623e84d99254204b2a1534e5d5) diff, I can see following change - probably made outside of this PR, but no idea where:
Screenshot 2021-07-14 at 14 47 30

Are you sure `||` operator is correct there? If yes, then I suspect following simplified version should work too:
```ts
const options = { where: filters, take, skip, cursor }
```

src/query/sortResults.ts
```diff
@@ -0,0 +1,89 @@
...
+): FlatSortCriteria[] {
+  log('flattenSortCriteria:', orderBy, propertyPath)
+
+  return orderBy.reduce<FlatSortCriteria[]>((criteria, properties) => {
+    warnOnIneffectiveSortingKeys(properties)
+
+    // Multiple properties in a single criteria object are forbidden.
+    // Use the list of criteria objects for multi-criteria sort.
+    const property = Object.keys(properties)[0] as keyof OrderBy<EntityType>
+    const sortDirection = properties[property]!
```
Aprillion, line 34: hm, I would avoid `!` here, it can be really `undefined` in runtime, e.g. if a user saves in the middle of editing:`[{}, {second: 'asc'}]`

Aprillion (approved):

kettanaito:
@Aprillion, yes `||` is correct there: if at least _one_ of `take` or `skip`/`cursor` is set, they must be added to the options. If I'm not mistaken, using `const options = { where: filters, take, skip, cursor }` is going to append them all the time?

kettanaito:
I'd like to wrap this up but there seem to be some typescript violations in the related code. Would appreciate any help available. It's enough to clone the repo, checkout this branch, and try to build it with the latest TypeScript.

fildon:
> I'd like to wrap this up but there seem to be some typescript violations in the related code. Would appreciate any help available. It's enough to clone the repo, checkout this branch, and try to build it with the latest TypeScript.

Just did it now. Build works locally for me :-/

kettanaito:
I've adjusted the typings test to pass in CI. The tests _are_ valid, it only looks like it validates the second wrong key/value pair and not the first.

PR #110, opened 2021-07-28
Supports relational types in GraphQL

### LINKED ISSUE #99: Feature Request/Unexpected behaviour: Graphql handlers just return string ids instead of relationships
I just noticed that over in #96 while adding tests.

If I were to add a second model "Post" with a `oneOf` relationship "author" to "User", that would still be output as `String` instead of referencing the `User` type, which would be expected of a graphql api.

This is not a problem of my other PR, but a conceptual problem with `toHandlers`. It would probably have to be part of `Factory` instead of being part of `Model`, to handle these spanning relationships and quite some additional work on `getGraphQLType`, which is very simple right now:

https://github.com/mswjs/data/blob/f8da857b3ea62dc2486b13f7cf9228e15e173b0d/src/model/generateGraphQLHandlers.ts#L35-L45

kettanaito:
- Closes #99

### DIFF
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -98,13 +98,15 @@ export type InternalEntity<
 
 export type ModelDictionary = Limit<Record<string, Record<string, any>>>
 
+export type ModelDictionaryValue<T extends Record<string, any>> =
+  | (() => BaseTypes)
+  | PrimaryKeyDeclaration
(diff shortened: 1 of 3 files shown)


kettanaito:
Request handler generation isn't a responsibility of this library anymore. It should be done in `@msw/source` instead.

PR #113, opened 2021-08-08
Nested objects

gidesan:


src/model/parseModelDefinition.ts
```diff
@@ -58,6 +60,14 @@ export function parseModelDefinition<Dictionary extends ModelDictionary>(
         return result
       }
 
+      if (isStrictlyObject(definition[property])) {
+        const flattenedProperties = Object.keys(
+          flatten(definition[property]),
```
kettanaito, line 65: Can we utilize [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap) here instead?

test/model/create.test.ts
```diff
@@ -29,3 +29,63 @@ test('creates a new entity with initial values', () => {
...
+
+  const exactUser = db.user.create({
+    id: 'abc-123',
+    arrayProp: [1, 2, 3],
+  })
+  expect(exactUser).toHaveProperty('id', 'abc-123')
+  expect(exactUser).toHaveProperty('arrayProp', [1, 2, 3])
+})
+
+test('creates a new entity with initial values, including nested objects', () => {
```
kettanaito, line 49:
```suggestion
test('supports nested objects in the model definition', () => {
```

test/model/create.test.ts
```diff
@@ -29,3 +29,63 @@ test('creates a new entity with initial values', () => {
...
+  const db = factory({
+    user: {
+      id: primaryKey(datatype.uuid),
+      name: String,
+      info: {
+        firstName: String,
```
kettanaito, line 55: We need a test case that ensures that relational properties can be declared in nested objects.

```js
{
  user: {
    billingAddress: {
      country: oneOf('country')
    }
  },
  country: { ... }
}
```

src/model/createModel.ts
```diff
@@ -38,40 +41,45 @@ export function createModel<
...
+      set(entity, property, propertyDefinition())
       return entity
-    },
+    }
+
+    return entity
+  }
+
+  const resolvedProperties = properties.reduce<Record<string, any>>(
+    resolveProperty,
```
kettanaito, line 82: Why separating the reducer into its own function? Is it a purely aesthetic change?

kettanaito (changes requested):
Hey, @gidesan. Thank you for working on the nested objects support in the models!

I've done the code review and have a few questions and clarifications that I'd appreciate you answering. Overall, I think the changes are good, I just need to get my ahead around flattenning the nested object keys and handling them like pointers. I wonder if there's a more efficient way to this, so we don't have to flatten potentially deep data structures.

test/model/create.test.ts
gidesan: wrote the test, it failed...working on it to make it pass :)

test/model/create.test.ts
```diff
@@ -29,3 +29,63 @@ test('creates a new entity with initial values', () => {
...
+        lastName: String,
+        address: {
+          street: () => 'Yellow Brick Road',
```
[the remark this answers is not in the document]
gidesan: fully agree, will improve tests and then feature code

src/model/createModel.ts
gidesan: I found it more readable, we can revert it if needed

src/utils/isInternalEntity.ts
```diff
@@ -8,6 +9,7 @@ export function isInternalEntity(
 ): value is InternalEntity<any, any> {
   return (
     value &&
+    isStrictlyObject(value) &&
```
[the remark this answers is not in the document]
gidesan: otherwise there are issues with array properties, which means the test `creates a new entity with initial values, including arrays` fails

src/model/parseModelDefinition.ts
```diff
@@ -58,6 +60,14 @@ export function parseModelDefinition<Dictionary extends ModelDictionary>(
         return result
       }
 
+      if (isStrictlyObject(definition[property])) {
```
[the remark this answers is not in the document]
gidesan: will check

test/model/create.test.ts
kettanaito: I believe you'll need to analyze model properties recursively because of nested objects support. Should be a matter of introducing recursion to `parseModelDeclaration` or friends.

src/utils/isInternalEntity.ts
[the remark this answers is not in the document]
kettanaito: Thank you for the explanation. I think we can drop this change and adopt #117 once it's merged. It does the same thing.

src/model/createModel.ts
kettanaito: I'd say that keeping line changes scoped to the feature is always a good idea. We may do changes like reducer functions separation in a standalone pull request aimed just at that. What are your thoughts on this?

src/model/parseModelDefinition.ts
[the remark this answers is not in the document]
kettanaito: In case it doesn't, please don't hurry to implement array support. I like the work done in #98, and I'd love to see both of your contributions gradually combined.

kettanaito:
If you update against the latest main branch, you'll get `isObject` assertion in the `isInternalEntity` and can remove the custom logic added here. Thanks!

src/model/createModel.ts
gidesan: That makes sense. Will revert it :)

test/model/create.test.ts
gidesan: @kettanaito the test is now passing (other ones are failing, currently having a look), altough the code can be further improved, cleaned, simplified.
Btw I pushed it to see if the direction is good :) 
I also got rid of `flat` library as you suggested.

test/model/create.test.ts
```diff
@@ -29,3 +29,63 @@ test('creates a new entity with initial values', () => {
...
+
+  // Entity can be given exact values to seed.
+  const exactUser = db.user.create({
+    id: 'abc-123',
+    name: 'sampleUser',
+    info: {
+      firstName: 'Reginald',
+      lastName: 'Dwight',
+      address: {
+        number: 73,
```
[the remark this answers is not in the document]
gidesan: Added tests for nested arrays (`if no initial values for nested arrays, the value getters will be used instead`), but not for nested objects. For nested object, the value getters used will be the ones related for "leaf" properties.

src/model/parseModelDefinition.ts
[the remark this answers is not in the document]
gidesan: it seem to work, for my expectations at least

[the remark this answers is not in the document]
kettanaito: Thanks for checking!

src/model/defineRelationalProperties.ts
```diff
@@ -107,17 +112,44 @@ export function defineRelationalProperties(
...
+        }
+        return acc[curr]
+      }, entity)
+    } else if (isObject(value)) {
+      defineNestedProperties(entity, value, nestedPath)
+    }
+  }
+}
+
+function isRelationalProperty(val: any) {
```
kettanaito, line 153: I recommend reusing the existing `isOneOfRelation` and `isManyOfRelation` functions from here:

https://github.com/mswjs/data/blob/3016d6d0e8bdff85995933a69f7162c1ff081734/src/utils/removeInternalProperties.ts#L10-L23

Do you think they'd cover your intentions here? If not, I'd like to learn more why (based on the function names, their intentions are the same).

src/model/parseModelDefinition.ts
```diff
@@ -21,8 +22,15 @@ export function parseModelDefinition<Dictionary extends ModelDictionary>(
   dictionary: Dictionary,
   modelName: string,
   definition: ModelDefinition,
+  propertyPath = '',
```
kettanaito, line 25: I don't think that exposing the internal details like `propertyPath` is a good idea in this function.

The intention behind `parseModelDefinition` is to produce a representation of the model given a dictionary. I understand the technical reasons you're adding this argument (to support recursion) but I think this is done at the wrong level. 

I'd much prefer for the `parseModelDefinition` function to remain as-is and solve the recursive parsing otherwise.

test/model/create.test.ts
```diff
@@ -29,3 +30,141 @@ test('creates a new entity with initial values', () => {
   })
   expect(exactUser).toHaveProperty('id', 'abc-123')
 })
+
+test('creates a new entity with initial values, including arrays', () => {
```
kettanaito, line 34: This test case is only about the arrays. I'd name it accordingly:

```js
test('creates a new entity with an array property')
```

test/model/create.test.ts
```diff
@@ -29,3 +30,141 @@ test('creates a new entity with initial values', () => {
...
+    },
+  })
+
+  expect(exactUser).toHaveProperty('name', 'user')
+  expect(exactUser.info).toHaveProperty('firstName', 'Ryuichi')
+  expect(exactUser.info).toHaveProperty('lastName', 'Sakamoto')
+  expect(exactUser.info.country).toHaveProperty('name', 'Japan')
+})
+
+test('throws an exception when primary key is not set at root level', () => {
```
kettanaito, line 130: May I please ask you to move any primaryKey-related test cases to `test/primaryKey.test.ts`?

It already contains some tests for primary key exceptions.

test/model/create.test.ts
```diff
@@ -29,3 +30,141 @@ test('creates a new entity with initial values', () => {
...
+      },
+    })
+  })
+  expect(error).toHaveProperty(
+    'message',
+    'Failed to parse a model definition for "info" property of "user": primary key cannot be in nested objects',
+  )
+})
+
+test('if no initial values for nested arrays, the value getters will be used instead', () => {
```
kettanaito, line 149:
```js
test('uses value getters when creating an entity with nested arrays')
```

kettanaito (changes requested):
Hey, @gidesan. Thank you once more for covering my suggestions!

I do have some concerns about some API changes in the parsing function, but I don't wish to block you while I think of a way to solve them. There are a few comments regarding moving the tests and reusing existing utilities that you may address if you have time.

kettanaito:
If you're comfortable with Git, we should clean up the commit history for this change. Each commit should represent a finite meaningful change. For example:

- chore: migrate to "lodash" package
- feat: support nested objects and arrays in the model definition
- ...

You can leave this formatting to me if you feel unsure. We can rebase the commits once the change is finished.

src/model/defineRelationalProperties.ts
gidesan: Unfortunately they do different things. Well, maybe similar but in different steps of the process.
Ofc I'm totally open to provide a different name for `isRelationalProperty` or implement it in a different way. But I wasn't able to make the suggested change work.

For example, in the `defineRelationalProperties` function with the following `entity` input (
```
    {
      id: '3ff81810-2161-457a-9b1a-2219a73db387',
      name: 'user',
      info: { firstName: 'Ryuichi', lastName: 'Sakamoto' },
      __type: 'user',
      __primaryKey: 'id'
    }
```
the `properties` object is constructed this way
```
properties: { country: { enumerable: true, get: [Function: get] } },
```
and, given that `Object.defineProperties` won't work for nested properties, I had to setup a `defineNestedProperties` function.

gidesan:
> If you're comfortable with Git, we should clean up the commit history for this change. Each commit should represent a finite meaningful change. For example:
> 
> * chore: migrate to "lodash" package
> * feat: support nested objects and arrays in the model definition
> * ...
> 
> You can leave this formatting to me if you feel unsure. We can rebase the commits once the change is finished.

Tbh I used not-so-meaningful commit messages thinking that we would have squashed them in the end. Even my email should be changed in the commits.
I could try to end the work, then generate a patch, then create a new branch and split the patch in meaningful commits with proper comments. A bit boring, but I can do that.

src/model/parseModelDefinition.ts
gidesan: done, hope it's better now :)

gidesan:
> Hey, @gidesan. Thank you once more for covering my suggestions!
> 
> I do have some concerns about some API changes in the parsing function, but I don't wish to block you while I think of a way to solve them. There are a few comments regarding moving the tests and reusing existing utilities that you may address if you have time.

I should have addressed your new suggestions, thanks a lot :)

jogelin:
Really impatient for this PR to be merged :) 

We would like to use this library to generate all factories for all of our mocks in cypress/jest testing and for localhost development too. However not having primitive arrays is blocking us for choosing this library :s

kettanaito:
## Primary key and Relations as classes

I've rewritten the internal implementation of the primary key and relations and those are now represented as classes (see my post above for the motivation to do so). No user-facing changes were introduced, apart from a new added invariant:

- When producing (resolving) relation for a non-existing model, or existing model that has no primary key (unlikely, would fail on the parsing stage), an exception will be thrown.

kettanaito:
Everything should be in order now: the feature is implemented, typed, relevant tests updated/added. 

Thank you for the great work on this, @gidesan! 🎉

PR #117, opened 2021-09-08
isInternalEntity: Assert the value is an object

### LINKED ISSUE #94: Handle Undefined Object in `isInternalEntity`
## Problem

When adding a non-entity in a relation property, the following error occurs: 

```
  TypeError: Cannot use 'in' operator to search for '__type' in null
```

This also breaks dropping a `db`.

## Expected Outcome

The code should not throw because of a null object, but error or warn the user about him using a normal object instead of an entity for a relation.

kettanaito:
## GitHub

- Fixes #94 
- A prerequisite for #98, #113 

## Changes

- Asserts that the given value is an object during the `isInternalEntity` check.

This prevents the `isInternalEntity` check to throw an exception when handling non-object types (still possible on runtime with the nested properties in model declarations).

```
TypeError: Cannot use 'in' operator to search for '__type' in "some value"
```

### DIFF
--- a/src/utils/isObject.ts
+++ b/src/utils/isObject.ts
@@ -1,0 +1,6 @@
+/**
+ * Returns true if the given value is a plain Object.
+ */
+export function isObject(value: any): value is Record<string, any> {
+  return value != null && typeof value === 'object' && !Array.isArray(value)
+}
--- a/src/utils/isInternalEntity.ts
+++ b/src/utils/isInternalEntity.ts
@@ -1,13 +1,14 @@
 import { InternalEntity, InternalEntityProperty } from '../glossary'
+import { isObject } from './isObject'
 
 /**
- * Determines if the given object is an internal entity.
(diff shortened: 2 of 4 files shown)


kettanaito:
Hey, @gidesan. Please, would you have a minute to review this change? I believe we can use it in your pull request, as well as in #98. I feel this is a general bug fix we should introduce regardless of the features. Thanks.

src/utils/isObject.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Returns true if the given value is a plain Object.
+ */
+export function isObject(value: any): value is Record<string, any> {
```
kettanaito, line 4: Using `unknown` is more suitable here, but there have been some issues with adopting `unknown` for our TypeScript users in the past. I'd migrate from `any` to `unknown`as a separate, carefully released change.

PR #141, opened 2021-10-22
fix(factory): return non-nullable data for "strict" queries

### LINKED ISSUE #140: Make the return type of query methods non-nullable when using the "strict" option
```ts
const user = db.user.findFirst({
  where: { ... },
  strict: true
})
```

The value of `user` cannot be null. It's either a value or an exception is thrown.

kettanaito:
- Fixes #140

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -125,7 +125,7 @@ function createModelApi<
         )
       }
 
-      return firstResult
+      return firstResult as Entity<any, any>
     },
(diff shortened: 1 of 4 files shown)


kettanaito:
But we can isolate `QueryOptions` into its own type and move it to the generic on query methods. That will allow the `strict` type inference without breaking the `QuerySelector` when it's moved into a generic entirely.

PR #143, opened 2021-10-22
Add support for null property values

### LINKED ISSUE #142: Support null value types
Hi 👋 

I'm loving the project, I've just finished integrating it into our gql mocking with msw. It's really simplified how we handle mutations so already a huge win!

I am hoping to add support for null property values. We use Apollo server, which represents missing values for optional types as null, so for a GQL type:

```gql
type Account {
  id: String!
  firstName: String
}
```
The resulting Account type would be represented as:
```typescript
type Account = {
  id: string;
  firstName: string | null;
}
```

Ideally I'd want to use a factory function to create realistic defaults, and then set certain properties to null when necessary:
```typescript
const db = factory({
  account: {
    id: primaryKey(String),
    firstName: () => faker.name.firstName() as string | null,
  }
})

const account = db.account.create()
const namelessAccount = db.account.create({
  firstName: null,
  lastName: null,
})
```
However I am unable to do that since null is not a valid return type from the factory functions. Instead I made sure the optional types always had a value. 
```typescript
const db = factory({
  account: {
    id: primaryKey(String),
    firstName: faker.name.firstName,
  }
})
```
This means our mock responses don't perfectly match our production responses because I can't initialise or update accounts to have a non-string firstName.

I will raise a PR with a naive implementation now, at least to clarify what I'm talking about.  Looking forwards to hearing what you think.

All the best, the project is awesome! 🙌

olivierwilkinson:
closes #142 

New `nullable` function provides an API for defining nullable properties
as well as nullable relations.

The arguments passed to `nullable` are identical to those passed to
factory when the property is not nullable. This means `nullable` should
feel natural to use for users familiar to the project.

For example to define a nullable property the user passes the factory
function they would use to normally define the property to `nullable`
when defining the db. This matches the `primaryKey` method as a bonus.

To define a nullable relation they would pass the `oneOf` or `manyOf`
function with the model name directly to `nullable`.

The model API methods are all typed to know when it's possible for a
property to be null, and an invariant violation is raised for js users
who try to update a non-nullable property to null.

### DIFF
--- a/src/model/createModel.ts
+++ b/src/model/createModel.ts
@@ -62,7 +62,7 @@ export function createModel<
         set(
           properties,
           propertyName,
-          initialValue || propertyDefinition.getValue(),
+          initialValue || propertyDefinition.getPrimaryKeyValue(),
         )
         return properties
       }
--- a/src/model/createModel.ts
+++ b/src/model/createModel.ts
@@ -1,4 +1,5 @@
 import { debug } from 'debug'
+import { invariant } from 'outvariant'
 import get from 'lodash/get'
 import set from 'lodash/set'
 import isFunction from 'lodash/isFunction'
@@ -75,6 +76,13 @@ export function createModel<
         set(properties, propertyName, value)
         return properties
       }
+
+      invariant(
+        initialValue !== null,
+        'Failed to create a "%s" entity: a non-nullable property "%s" cannot be instantiated with null. Use the "nullable" function when defining this property to support nullable value.',
+        modelName,
+        propertyName.join('.'),
+      )
 
       if (isModelValueType(initialValue)) {
         log(
--- a/src/model/createModel.ts
+++ b/src/model/createModel.ts
@@ -16,6 +16,8 @@ import { ParsedModelDefinition } from './parseModelDefinition'
 import { defineRelationalProperties } from './defineRelationalProperties'
 import { PrimaryKey } from '../primaryKey'
 import { Relation } from '../relations/Relation'
+import { NullableProperty } from '../nullable'
+import { isModelValueType } from '../utils/isModelValueType'
 
 const log = debug('createModel')
 
@@ -64,14 +66,17 @@ export function createModel<
         return properties
       }
 
-      if (
-        typeof initialValue === 'string' ||
-        typeof initialValue === 'number' ||
-        typeof initialValue === 'boolean' ||
-        // @ts-ignore
-        initialValue?.constructor.name === 'Date' ||
-        Array.isArray(initialValue)
-      ) {
+      if (propertyDefinition instanceof NullableProperty) {
+        const value =
+          initialValue === null || isModelValueType(initialValue)
+            ? initialValue
+            : propertyDefinition.getValue()
+
+        set(properties, propertyName, value)
+        return properties
+      }
+
+      if (isModelValueType(initialValue)) {
         log(
           '"%s" has a plain initial value:',
           `${modelName}.${propertyName}`,
--- a/src/model/defineRelationalProperties.ts
+++ b/src/model/defineRelationalProperties.ts
@@ -26,9 +26,17 @@ export function defineRelationalProperties(
       relation.target.modelName,
     )
 
-    const references: Value<any, ModelDictionary> | undefined = get(
+    const references: Value<any, ModelDictionary> | null | undefined = get(
       initialValues,
       propertyPath,
+    )
+
+    invariant(
+      references !== null || relation.attributes.nullable,
+      'Failed to define a "%s" relational property to "%s" on "%s": a non-nullable relation cannot be instantiated with null. Use the "nullable" function when defining this relation to support nullable value.',
+      relation.kind,
+      propertyPath.join('.'),
+      entity[ENTITY_TYPE],
     )
 
     log(
--- a/src/model/defineRelationalProperties.ts
+++ b/src/model/defineRelationalProperties.ts
@@ -32,7 +32,9 @@ export function defineRelationalProperties(
     )
 
     log(
-      `setting relational property "${entity.__type}.${propertyPath.join('.')}" with references: %j`,
+      `setting relational property "${entity.__type}.${propertyPath.join(
+        '.',
+      )}" with references: %j`,
       relation,
       references,
     )
@@ -41,7 +43,7 @@ export function defineRelationalProperties(
 
     if (references) {
       relation.resolveWith(entity, references)
-    } else if (relation.nullable) {
+    } else if (relation.attributes.nullable) {
       relation.resolveWith(entity, null)
     }
   }
--- a/src/model/defineRelationalProperties.ts
+++ b/src/model/defineRelationalProperties.ts
@@ -41,6 +41,8 @@ export function defineRelationalProperties(
 
     if (references) {
       relation.resolveWith(entity, references)
+    } else if (relation.nullable) {
+      relation.resolveWith(entity, null)
     }
   }
 }
--- a/src/model/parseModelDefinition.ts
+++ b/src/model/parseModelDefinition.ts
@@ -9,6 +9,7 @@ import {
 import { PrimaryKey } from '../primaryKey'
 import { isObject } from '../utils/isObject'
 import { Relation, RelationsList } from '../relations/Relation'
+import { NullableProperty } from '../nullable'
 
 const log = debug('parseModelDefinition')
 
@@ -66,6 +67,12 @@ function deepParseModelDefinition<Dictionary extends ModelDictionary>(
       result.primaryKey = propertyName
       result.properties.push([propertyName])
 
+      continue
+    }
+
+    if (value instanceof NullableProperty) {
+      // Add nullable properties to the same list as regular properties
+      result.properties.push(propertyPath)
       continue
     }
 
--- a/src/nullable.ts
+++ b/src/nullable.ts
@@ -16,16 +16,20 @@ export function nullable<ValueType extends ModelValueType>(
   value: NullableGetter<ValueType>,
 ): NullableProperty<ValueType>
 
-export function nullable<ValueType extends Relation<any, any, any, false>>(
+export function nullable<
+  ValueType extends Relation<any, any, any, { nullable: false }>,
+>(
   value: ValueType,
-): ValueType extends Relation<infer Kind, infer Key, any, false>
+): ValueType extends Relation<infer Kind, infer Key, any, { nullable: false }>
   ? Kind extends RelationKind.ManyOf
     ? ManyOf<Key, true>
     : OneOf<Key, true>
   : never
 
 export function nullable(
-  value: NullableGetter<ModelValueType> | Relation<any, any, any, false>,
+  value:
+    | NullableGetter<ModelValueType>
+    | Relation<any, any, any, { nullable: false }>,
 ) {
   if (typeof value === 'function') {
     return new NullableProperty(value)
@@ -33,8 +37,10 @@ export function nullable(
 
   return new Relation({
     kind: value.kind,
-    attributes: value.attributes,
     to: value.target.modelName,
-    nullable: true,
+    attributes: {
+      ...value.attributes,
+      nullable: true,
+    },
   })
 }
--- a/src/nullable.ts
+++ b/src/nullable.ts
@@ -1,0 +1,40 @@
+import { ModelValueType } from './glossary'
+import { ManyOf, OneOf, Relation, RelationKind } from './relations/Relation'
+
+export type NullableGetter<ValueType extends ModelValueType> =
+  () => ValueType | null
+
+export class NullableProperty<ValueType extends ModelValueType> {
+  public getValue: NullableGetter<ValueType>
+
+  constructor(getter: NullableGetter<ValueType>) {
+    this.getValue = getter
+  }
+}
+
+export function nullable<ValueType extends ModelValueType>(
+  value: NullableGetter<ValueType>,
+): NullableProperty<ValueType>
+
+export function nullable<ValueType extends Relation<any, any, any, false>>(
+  value: ValueType,
+): ValueType extends Relation<infer Kind, infer Key, any, false>
+  ? Kind extends RelationKind.ManyOf
+    ? ManyOf<Key, true>
+    : OneOf<Key, true>
+  : never
+
+export function nullable(
+  value: NullableGetter<ModelValueType> | Relation<any, any, any, false>,
+) {
+  if (typeof value === 'function') {
+    return new NullableProperty(value)
+  }
+
+  return new Relation({
+    kind: value.kind,
+    attributes: value.attributes,
+    to: value.target.modelName,
+    nullable: true,
+  })
+}
--- a/src/query/compileQuery.ts
+++ b/src/query/compileQuery.ts
@@ -34,7 +34,7 @@ export function compileQuery<Data extends Record<string, any>>(
 
         // If an entity doesn't have any value for the property
         // is being queried for, treat it as non-matching.
-        if (typeof actualValue === 'undefined') {
+        if (actualValue == null) {
           return false
         }
 
--- a/src/utils/isModelValueType.ts
+++ b/src/utils/isModelValueType.ts
@@ -1,0 +1,17 @@
+import { ModelValueType, PrimitiveValueType } from '../glossary'
+
+function isPrimitiveValueType(value: any): value is PrimitiveValueType {
+  return (
+    typeof value === 'string' ||
+    typeof value === 'number' ||
+    typeof value === 'boolean' ||
+    value?.constructor?.name === 'Date'
+  )
+}
+
+export function isModelValueType(value: any): value is ModelValueType {
+  return (
+    isPrimitiveValueType(value) ||
+    (Array.isArray(value) && value.every(isPrimitiveValueType))
+  )
+}
--- a/test/model/create.test-d.ts
+++ b/test/model/create.test-d.ts
@@ -14,8 +14,8 @@ const db = factory({
     },
   },
   company: {
-    name: primaryKey(String)
-  }
+    name: primaryKey(String),
+  },
 })
 
 // @ts-expect-error Unknown model name.
@@ -30,6 +30,11 @@ db.user.create({
       country: 'us',
     },
   },
+})
+
+db.user.create({
+  // @ts-expect-error Non-nullable properties cannot be instantiated with null.
+  firstName: null,
 })
 
 db.user.create({
--- a/test/model/create.test-d.ts
+++ b/test/model/create.test-d.ts
@@ -1,15 +1,21 @@
-import { factory, primaryKey } from '@mswjs/data'
+import { factory, primaryKey, nullable } from '@mswjs/data'
+import faker from 'faker'
 
 const db = factory({
   user: {
     id: primaryKey(String),
     firstName: String,
+    lastName: nullable(faker.name.lastName),
+    age: nullable<number>(() => null),
     address: {
       billing: {
         country: String,
       },
     },
   },
+  company: {
+    name: primaryKey(String)
+  }
 })
 
 // @ts-expect-error Unknown model name.
@@ -60,3 +66,15 @@ db.user.create({
   // will be executed to get the initial value.
   firstName: 'John',
 })
+
+const user = db.user.create({
+  // Nullable properties can have an initialValue of null or the property type
+  lastName: null,
+  age: 15,
+})
+
+// @ts-expect-error lastName property is possibly null
+user.lastName.toUpperCase()
+
+// @ts-expect-error property 'toUpperCase' does not exist on type 'number'
+user.age?.toUpperCase()
--- a/test/model/create.test.ts
+++ b/test/model/create.test.ts
@@ -1,5 +1,5 @@
-import { datatype } from 'faker'
-import { factory, primaryKey, oneOf } from '@mswjs/data'
+import { datatype, name } from 'faker'
+import { factory, primaryKey, oneOf, nullable } from '@mswjs/data'
 import { identity } from '../../src/utils/identity'
 
 test('creates a new entity', () => {
@@ -44,6 +44,29 @@ test('creates a new entity with an array property', () => {
   })
   expect(exactUser).toHaveProperty('id', 'abc-123')
   expect(exactUser).toHaveProperty('arrayProp', [1, 2, 3])
+})
+
+test('creates a new entity with nullable properties', () => {
+  const db = factory({
+    user: {
+      id: primaryKey(datatype.uuid),
+      name: nullable(name.findName),
+      age: nullable<number>(() => null),
+      address: {
+        street: String,
+        number: nullable<number>(() => null),
+      },
+    },
+  })
+
+  const user = db.user.create({
+    id: 'abc-123',
+    name: null,
+  })
+
+  expect(user).toHaveProperty('name', null)
+  expect(user).toHaveProperty('age', null)
+  expect(user.address).toHaveProperty('number', null)
 })
 
 test('supports nested objects in the model definition', () => {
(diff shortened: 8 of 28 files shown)


src/model/createModel.ts
```diff
@@ -51,7 +53,10 @@ export function createModel<
       const propertyDefinition = get(definition, propertyName)
 
       // Ignore relational properties at this stage.
-      if (propertyDefinition instanceof Relation) {
+      if (
+        propertyDefinition instanceof Relation ||
+        propertyDefinition instanceof NullableRelation
```
kettanaito, line 58: Shall we consider implementing `NullableRelation` as a child of the `Relation` base class?

That way we don't have to have any explicit `instanceof NullableRelation` checks.

```js
import { Relation } from './relation'

export class NullableRelation extends Relation {
  // Add nullable values support
}

const relation = new NullableRelation()
if (relation instanceof Relation) {
  // still true
}
```

kettanaito: With inheritance, we can utilize parent class checks when we intend to handle `Relation` and `NullableRelation` the same, and strict `instanceof NullableRelation` when we wish to implement the logic specific only to the nullable relation. 

What do you think about this?

src/model/parseModelDefinition.ts
```diff
@@ -69,13 +70,25 @@ function deepParseModelDefinition<Dictionary extends ModelDictionary>(
       continue
     }
 
+    if (value instanceof NullableProperty) {
+      // Add nullable properties to the same list as regular properties
+      result.properties.push(propertyPath)
```
kettanaito, line 75: Once again, a great example of how extending the `Relation` class for `NullableRelation` would allow us to remove this branch altogether.

src/model/parseModelDefinition.ts
```diff
@@ -69,13 +70,25 @@ function deepParseModelDefinition<Dictionary extends ModelDictionary>(
       continue
     }
 
+    if (value instanceof NullableProperty) {
```
kettanaito, line 73: Won't it have the same effect if we omit this `if` altogether? It should fall down to the very last `result.properties.push()`, doing the same thing. Let me know if I'm missing something here.

src/nullable.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+export function nullable<ValueType extends ModelValueType>(
+  value: NullableGetter<ValueType>,
+): NullableProperty<ValueType>
+export function nullable<RelationType extends Relation<any, any, any>>(
+  value: RelationType,
+): NullableRelation<RelationType>
+export function nullable(value: any) {
```
kettanaito, line 29: I believe you don't have to annotate the `value` here, it should be inferred from the overloads above. Isn't that the case?

src/query/compileQuery.ts
```diff
@@ -34,7 +34,7 @@ export function compileQuery<Data extends Record<string, any>>(
 
         // If an entity doesn't have any value for the property
         // is being queried for, treat it as non-matching.
-        if (typeof actualValue === 'undefined') {
+        if (typeof actualValue === 'undefined' || actualValue === null) {
```
kettanaito, line 37: We can shorten this to `actualValue == null` then. It produces the same result as these two conditions above.

src/nullable.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor(getter: NullableGetter<ValueType>) {
+    this.getValue = getter
+  }
+}
+
+export class NullableRelation<RelationType extends Relation<any, any, any>> {
+  public relation: RelationType
+
+  constructor(relation: RelationType) {
+    this.relation = relation
```
kettanaito, line 19: I think we're extending the base `Relation` class incorrectly for this use case. 

Instead of adding a new `resolveWithNull` method to the base class, override the `resolveWith` method on the `NullableRelation` class. 

> `Relation` on its own doesn't care/know anything about the nullability. It's by the introduction of a child `NullableRelation` class you're creating a subset of the `Relation` that can now handle nullable values. 

```ts
export class NullableRelation extends Relation {
  resolveWith(...args) {
    // Support nullable next values.
  }
}
```

With this change, we can also handle `Relation` and `NullableRelation` instances _the same way_because we don't introduce any custom methods like `resolveWithNull`. Instead, we delegate the nullable support to the child class, keeping the same interface as the parent (which makes sense: when using `NullableRelation` expect `resolveWith` to support nullable next values).

kettanaito: **Alternatively**, we can just add the nullable values support straight to the `Relation` class and don't complicate things with the designated `NullableRelation` child. This would largely depend on how complicated the `resolveWihNull` logic turns out to be. If it's as simple as checking if the next value is `null` and allowing that, I'd vote for changing the `Relation` class instead.

test/model/create.test-d.ts
```diff
@@ -60,3 +63,14 @@ db.user.create({
...
+const user = db.user.create({
+  // It is possible to create models with null values
+  lastName: null,
+})
+
+// @ts-expect-error lastName property is possibly null
+user.lastName.toUpperCase()
+
+// @ts-expect-error property 'toUpperCase' does not exist on type 'number'
+user.age?.toUpperCase()
```
kettanaito, line 76: Shouldn't TypeScript complain about the property being possibly null first? 🤔

test/model/create.test.ts
```diff
@@ -148,3 +148,24 @@ test('uses value getters when creating an entity with nested arrays', () => {
   expect(exactUser.info).toHaveProperty('tags', [1, 2])
   expect(exactUser.info).toHaveProperty('documents', [])
 })
+
+test('can create models with null values', () => {
```
kettanaito, line 152:
```suggestion
test('creates a new entity with nullable properties', () => {
```

test/model/create.test.ts
```diff
@@ -148,3 +148,24 @@ test('uses value getters when creating an entity with nested arrays', () => {
...
+      address: {
+        street: String,
+        number: nullable<number>(() => null),
+      }
+    },
+  })
+
+  const user = db.user.create({
+    id: 'abc-123',
+    name: null,
```
kettanaito, line 166: Do I understand correctly that here we're checking that the nullable property with an explicit initial value of `null` preserves that value? 

If so, we should add another case for the root-level nullable property _without_ the initial value in `.create()` to make sure it acquires the `null` value implicitly.

test/model/update.test-d.ts
```diff
@@ -42,6 +45,9 @@ db.user.update({
     firstName: 'next',
     age: 24,
     country: db.country.create({ code: 'de' }),
+    lastName: null,
+    // @ts-expect-error Unable to update non-nullable values to null
+    updatedAt: null
```
kettanaito, line 50: _This_ is gorgeous.

test/model/update.test.ts
```diff
@@ -510,3 +510,72 @@ test('exposes a root entity for a derivitive value of a nested property', () =>
 
   expect(result).toHaveProperty(['address', 'billing', 'country'], 'US')
 })
+
+test('sets properties to and from null', () => {
```
kettanaito, line 514:
```suggestion
test('supports updating nullable properties', () => {
```

test/model/update.test.ts
```diff
@@ -510,3 +510,72 @@ test('exposes a root entity for a derivitive value of a nested property', () =>
...
+
+  expect(() =>
+    db.user.update({
+      where: {
+        id: {
+          equals: 'abc-123',
+        },
+      },
+      data: {
+        // @ts-ignore
```
kettanaito, line 574: I'd generally prefer `@ts-expect-error` with a short description (like you did in one of the files above).

test/model/updateMany.test.ts
```diff
@@ -177,6 +177,58 @@ test('should update many entities with primitive values', () => {
   updateMultiUsers.forEach((user) => expect(user.role).toEqual('Admin'))
 })
 
+test('should update many entities to and from null values', () => {
```
kettanaito, line 180: The same comment here about splitting this into two separate test cases.

test/model/updateMany.test.ts
```diff
@@ -177,6 +177,58 @@ test('should update many entities with primitive values', () => {
...
+      role: {
+        equals: 'Auditor',
+      },
+    },
+    data: {
+      role: null,
+    },
+  })!
+
+  expect(oldAuditors).toHaveLength(2)
```
kettanaito, line 214: Using a `prev` prefix is a nice way to denote a previous state of a value. [See the naming cheatsheet](https://github.com/kettanaito/naming-cheatsheet#prevnext). 


```diff
-const oldAuditors = value
+const prevAuditors = value
```

Unless you mean "old of age" when saying "old" of course :D

test/model/updateMany.test.ts
```diff
@@ -177,6 +177,58 @@ test('should update many entities with primitive values', () => {
...
+  oldAuditors.forEach((user) => expect(user.role).toBeNull())
+
+  const newAdmins = db.user.updateMany({
```
kettanaito, line 217: A similar comment, `next` is a much better prefix for this case.

src/utils/isModelValueType.ts
```diff
@@ -0,0 +1,11 @@
+import { ModelValueType } from '../glossary'
+
+export function isModelValueType(value: any): value is ModelValueType {
```
kettanaito, line 3: What is the benefit of abstracting this? Do we use this check often?

src/model/defineRelationalProperties.ts
```diff
@@ -41,6 +48,8 @@ export function defineRelationalProperties(
 
     if (references) {
       relation.resolveWith(entity, references)
+    } else if (definition instanceof NullableRelation) {
+      relation.resolveWithNull(entity)
```
kettanaito, line 52: I think we should extend the `Relation` class to override its `resolveWith` method to support `null` as `references (thus preserving the call signature and getting rid of this condition). It troubles me that we'd have to treat `NullableRelation` as a special case everywhere we go. This can be eliminated by proper inheritance.

src/model/createModel.ts
```diff
@@ -64,14 +69,18 @@ export function createModel<
...
 
-      if (
-        typeof initialValue === 'string' ||
-        typeof initialValue === 'number' ||
-        typeof initialValue === 'boolean' ||
-        // @ts-ignore
-        initialValue?.constructor.name === 'Date' ||
-        Array.isArray(initialValue)
-      ) {
+      if (propertyDefinition instanceof NullableProperty) {
```
kettanaito, line 72: Why are we checking for nullable properties first? Would some of them fall under `isModelValueType` check below?

src/utils/isModelValueType.ts
kettanaito: Okay, I see. We're reusing it in `createModel` within the nullable property check.

src/model/createModel.ts
```diff
@@ -64,14 +69,18 @@ export function createModel<
...
+        set(
+          properties,
+          propertyName,
+          initialValue === null || isModelValueType(initialValue)
```
kettanaito, line 76: Do you mind extracting this value expression into a variable? It'd read much nicer.

```ts
const value = initialValue === null || isModelValueType(initialValue) ? initialValue : propertyDefinition.getValue()
set(properties, propertyName, value)
```

kettanaito (changes requested):
Hey, @olivierwilkinson. 

Your work on this change is outstanding. Thank you for improving Data! 👏 

I've left a few comments and would appreciate your feedback on those. My main concern goes to the proper extension of the `Relation` parent class in `NullableRelation`. I think we can lift off a few changes just by doing that.

kettanaito:
There's also a failing test (both locally and on CI), which suggest a regression:

```bash
FAIL test/primaryKey.test.ts
  ● Test suite failed to run

    test/primaryKey.test.ts:138:11 - error TS2578: Unused '@ts-expect-error' directive.

    138           // @ts-expect-error Primary key on nested properties are forbidden.
```

This may not be related to the changes directly. You can try resolving this, or we can look into it together after the comments above are addressed.

olivierwilkinson:
> Hey, @olivierwilkinson.
> 
> Your work on this change is outstanding. Thank you for improving Data! 👏
> 
> I've left a few comments and would appreciate your feedback on those. My main concern goes to the proper extension of the `Relation` parent class in `NullableRelation`. I think we can lift off a few changes just by doing that.

Hi @kettanaito,

Thanks for all the feedback! Rushing out the door right now but I will get on addressing your comments tomorrow evening with some luck 😄.

I really like your ideas on how to simplify `NullableRelation`; I'll compare and contrast inheriting from the `Relation` class versus adding nullable value support to it and report back 👍.

src/model/parseModelDefinition.ts
olivierwilkinson: The NullableProperty class gets caught by the `isObject` util since `typeof value === 'object'` where value is an instance of `NullableProperty` returns true. I could've written it as follows though:

```typescript
if (isObject<NestedModelDefinition>(value) && !(value instanceof NullableProperty)) {
  ...
}
```

The above is definitely clearer as to why it needs to be there but I kinda liked that `NullableProperty` was being handled explicitly which is why I went with what's there now. I'm very happy with either so up to you what you prefer 😄 .

There may be alternatives around changing the `isObject` util to be more specific but I don't know the implications of that.

src/nullable.ts
olivierwilkinson: It isn't inferred unfortunately but I did manage to improve the type by using a union of the different value types. I feel like typescript should've handled that for me like you suggest haha. Ah well 🤷‍♂️

test/model/create.test-d.ts
olivierwilkinson: Not in this case because we are using optional chaining 👍

test/model/create.test.ts
olivierwilkinson: Nice, yeah that reads better.

test/model/create.test.ts
olivierwilkinson: Yeah that is correct, it's checking that explicit initial values of `null` are preserved. Makes sense to me, I'll add the extra case 😁

test/model/update.test-d.ts
olivierwilkinson: I've realised from this comment that I need to add a check that initial values passed to `create` are also correctly typed 😄

test/model/updateMany.test.ts
olivierwilkinson: Oh I hadn't seen that naming cheatsheet, I'll have a read of it now 👍

src/utils/isModelValueType.ts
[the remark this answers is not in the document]
olivierwilkinson: Hmm, when I do that I get the following error message: 
`A type predicate's type must be assignable to its parameter's type.`

I think I can tweak the logic to make it work for any value, then the type is valid as `any`. What do you think about this? 

```typescript
function isPrimitiveValueType(value: any): value is PrimitiveValueType {
  return (
    typeof value === 'string' ||
    typeof value === 'number' ||
    typeof value === 'boolean' ||
    value?.constructor?.name === 'Date'
  )
}

export function isModelValueType(value: any): value is ModelValueType {
  return (
    isPrimitiveValueType(value) ||
    (Array.isArray(value) && value.every(isPrimitiveValueType))
  )
}
```

I can't think of a case where this would cause a false positive but I could be missing something.

src/model/createModel.ts
olivierwilkinson: I don't think it would be caught by the `isModelValueType` check, I put it there because it felt happy next to the `instanceof PrimaryKey` check haha. I can move it below `isModelValueType` though if you'd prefer?

test/model/update.test-d.ts
kettanaito: Great catch! 👍

test/model/create.test-d.ts
kettanaito: Oh, I see. Thanks for the explanation.

olivierwilkinson:
Hi @kettanaito,

I've updated the branch with a new implementation. I'm happy to say that both were an improvement on the previous implementation so was torn on which to go for. In the end I made whether the relation is nullable or a not a property of the `Relation` class itself as it seems more organised, however it required a perhaps unexpected use of Generics to make it possible to infer whether the relation is nullable or not.

I did also implement the other option of properly inheriting from `Relation` in `NullableRelation` which [you can have a look at here](https://github.com/olivierwilkinson/data/compare/main...null-value-support-inheritence) to see if you prefer it 😄. I would be very happy to move over to this solution if you felt it is better.

Thanks for your comments, they really improved the PR 🙌

olivierwilkinson:
I think I've fixed the failing `primaryKey.test.ts` suite now. Although I thought I had before so I'm going to wait for the next build to pass before uncrossing my fingers 🤞.

kettanaito:
I think there's a TypeScript version mismatch between the local environment and the CI.

kettanaito:
Hey, @olivierwilkinson 👋 

I've pushed a couple of changes to your remote:

1. `nullable` is now treated as a relation's attribute. Alongside `unique`, `nullable` is a property that modifies the relation's behavior. Such properties are defined as "attributes" in the library. I've adjusted the type generics to reflect that.
2. Added the relational value assertions to the tests where we update nullable relational values (in both success and failure scenarios). We need to be certain that the actual relational value is set to null or preserved when the relation is nullable and not, respectively. 

The build and tests pass on local, so I'll switch to look into this CI issue.

kettanaito:
I've updated the feature branch so now CI will report the versions of different tools (we're mainly interested in `typescript` and `tsc` compatibility). Let's see.

olivierwilkinson:
> Hey, @olivierwilkinson 👋
> 
> I've pushed a couple of changes to your remote:
> 
>     1. `nullable` is now treated as a relation's attribute. Alongside `unique`, `nullable` is a property that modifies the relation's behavior. Such properties are defined as "attributes" in the library. I've adjusted the type generics to reflect that.
> 
>     2. Added the relational value assertions to the tests where we update nullable relational values (in both success and failure scenarios). We need to be certain that the actual relational value is set to null or preserved when the relation is nullable and not, respectively.
> 
> 
> The build and tests pass on local, so I'll switch to look into this CI issue.

Very nice! Makes more sense for nullable to live in attributes looking at it now. 😄

Thanks for looking into the CI issue 🙌 I wasn't sure how I was going to fix that

src/nullable.ts
kettanaito: We resolved this by adding `refs === null` support to the original `Relation` class (its `resolveWith()` method).

test/model/update.test.ts
kettanaito: The tests were updated.

src/model/defineRelationalProperties.ts
kettanaito: The comment is obsolete, the `Relation` class now supports `refs` as `null`.

kettanaito:
I think the CI failure is related to the `typescript` and `tsc` versions mismatch. Here's what CI reports:

```
node: v14.18.1
npm: 6.14.15
typescript: @mswjs/data@0.7.2 /home/runner/work/data/data
└── typescript@4.3.5 
tsc: Version 4.4.4
```

While locally I have:

```
typescript: @mswjs/data@0.7.2 /mswjs/data
└── typescript@4.3.5
tsc: Version 4.3.5
```

Looks like CI is installing the latest `tsc` which has backward-incompatible changes with `4.3.5` (validates types differently).

kettanaito:
@olivierwilkinson, perhaps a silly question, but do we have any runtime checks that a non-nullable property _can't_ be set to null when calling `.create`/`.update`? Like throwing an exception.

olivierwilkinson:
> @olivierwilkinson, perhaps a silly question, but do we have any runtime checks that a non-nullable property _can't_ be set to null when calling `.create`/`.update`? Like throwing an exception.

Not a silly question at all! We have invariant violations when trying to update a non-nullable property or relation to null, but not for creation. Would you like me to add some for creation?

kettanaito:
If you have a chance, could you please add a few test cases to the existing `create.test.ts` that it throws an exception when creating an entity with its non-nullable property set to null? 

I think we should be explicit about this, so if the user wants to use `null` as a value, they must always mark that property as `nullable`.

olivierwilkinson:
> If you have a chance, could you please add a few test cases to the existing `create.test.ts` that it throws an exception when creating an entity with its non-nullable property set to null?
> 
> I think we should be explicit about this, so if the user wants to use `null` as a value, they must always mark that property as `nullable`.
> 
> P.S. Don't forget to pull the changes first ;)

Haha I won't squash either don't worry 😆

That is all done! 😄

kettanaito:
Thank you for adding the logic and tests when creating non-nullable properties with null. I've pushed to your branch, adjusting the wording, if you don't mind. It's beneficial to contain the current state in the error message ("a non-nullable property/relation") as well as explain why using "nullable" is necessary (and not necessarily what the user wants).

The changes look fantastic. I'll look into the failing CI and try to align the TS versions so the build is green.

kettanaito:
## Issue: `PrimaryKey` falls under `NullableProperty`

It seems that TypeScript is comparing class instances based on their properties/methods. This causes the `PrimaryKey` instances to match the `NullableProperty` class (both classes share `this.getValue` property). This results in primary keys being allowed in nested objects when defining a model, which is a no-op.

One way to solve that is to name those internal properties differently. I'm considering renaming `getValue` to `getPrimaryKeyValue` for the `PrimaryKey` class to prevent false positive matches then type-validating.

kettanaito:
Okay, so it wasn't a TS versions mismatch, in the end, it was a regression in primary key check. Once I pushed the fix, the CI is back to normal now.

kettanaito (approved):
A hard and definite "approved". You've done an incredible job bringing nullable support to Data! Thank you, @olivierwilkinson, and welcome to contributors! 🎉 

We are always looking for collaborators, so if any other issue in Data or MSW ecosystem libraries interests you, do not hesitate to discuss and dive into them. I know we could use your expertise.

olivierwilkinson:
> Thank you for adding the logic and tests when creating non-nullable properties with null. I've pushed to your branch, adjusting the wording, if you don't mind. It's beneficial to contain the current state in the error message ("a non-nullable property/relation") as well as explain why using "nullable" is necessary (and not necessarily what the user wants).

Great! No worries at all, I was struggling to figure out a concise way of phrasing it and I like what you changed it to 😄

olivierwilkinson:
> ## Issue: `PrimaryKey` instance matches the `NullableProperty` class
> 
> It seems that TypeScript is comparing class instances based on their properties/methods. This causes the `PrimaryKey` instances to match the `NullableProperty` class (both classes share `this.getValue` property). This results in primary keys being allowed in nested objects when defining a model, which is a no-op.
> 
> One way to solve that is to name those internal properties differently. I'm considering renaming `getValue` to `getPrimaryKeyValue` for the `PrimaryKey` class to prevent false positive matches then type-validating.

Oh wow it hadn't occured to me that's how they are comparing them, good to keep in mind! It makes sense I suppose since before your change both `NullableProperty` and `PrimaryKey` could be described by the same interface so a `PrimaryKey` instance does indeed extend `NullableProperty`. Nicely found 🙌

olivierwilkinson:
> A hard and definite "approved". You've done an incredible job bringing nullable support to Data! Thank you, @olivierwilkinson, and welcome to contributors! 🎉
> 
> We are always looking for collaborators, so if any other issue in Data or MSW ecosystem libraries interests you, do not hesitate to discuss and dive into them. I know we could use your expertise.

Amazing 🥳

Ah I'm happy to hear that! I'm keenly interested so I would love to continue contributing where I am able to. 

Thank you for putting so much of your own time and effort into reviewing and contributing to make it so polished. I'm really happy with the end result so I appreciate it a lot. This has been a pleasure! 😁

I'm going to go make some nullable properties now 💥

kettanaito:
All thanks go to you, really. It's by the power of great contributions that we can deliver quality software, together. Thank you!

PR #152, opened 2021-10-29
Support model properties with dots in them

### LINKED ISSUE #149: Properties with dots in their name are treated as paths instead of as literals
Given the following code:

```ts
export const db = factory({
  user: {
    'employee.id': primaryKey(() => 'abc-123'),
  }
})

db.user.create({ 'employee.id': 'some-id' })
```

We get this error:
```
Uncaught (in promise) OperationError: Failed to create a "user" entity: expected the primary key "employee.id" to have a value, but got: undefined
    at new OperationError (OperationError.js:28)
    at Object.create (factory.js:66)
```

This is because createModel uses `lodash/set`, which converts dot strings to deep properties (eg, among others):
https://github.com/mswjs/data/blob/c4a5208301660f4abcfa9996ce301c34f5047aa0/src/model/createModel.ts#L58-L65

The resulting model object is:
```ts
  const user: {
    employee: {
      id: <primaryKey>
    }
  }
  // instead of 
  const user: {
    'employee.id': <primaryKey>
  }
```

I think that `lodash/set` is a poor fit, because properties with dots in them are completely valid. Is there any reason it was used other than convenience?

VanTanev:
This PR moves the internal representation of property paths from dot notation strings to arrays of strings.
This allows us to continue using and benefitting from `lodash/set`, and also support dots in model properties, eg:

```ts
export const db = factory({
  user: {
    'employee.id': primaryKey(() => 'abc-123'),
  }
})
```

Closes #149

### DIFF
--- a/src/relations/Relation.ts
+++ b/src/relations/Relation.ts
@@ -67,7 +67,7 @@ export type ManyOf<ModelName extends KeyType> = Relation<
 >
 
 export type RelationsList = Array<{
-  path: string[]
+  propertyPath: string[]
   relation: Relation<any, any, any>
 }>
 
--- a/src/relations/Relation.ts
+++ b/src/relations/Relation.ts
@@ -32,7 +32,7 @@ export interface RelationAttributes {
 export interface RelationSource {
   modelName: string
   primaryKey: PrimaryKeyType
-  propertyPath: string
+  propertyPath: string[]
 }
 
 export interface RelationDefinition<
@@ -66,7 +66,10 @@ export type ManyOf<ModelName extends KeyType> = Relation<
   any
 >
 
-export type RelationsMap = Record<string, Relation<any, any, any>>
+export type RelationsList = Array<{
+  path: string[]
+  relation: Relation<any, any, any>
+}>
 
 const DEFAULT_RELATION_ATTRIBUTES: RelationAttributes = {
   unique: false,
@@ -118,7 +121,7 @@ export class Relation<
    */
   public apply(
     entity: Entity<any, any>,
-    propertyPath: string,
+    propertyPath: string[],
     dictionary: Dictionary,
     db: Database<Dictionary>,
   ) {
--- a/test/model/create.test.ts
+++ b/test/model/create.test.ts
@@ -148,3 +148,17 @@ test('uses value getters when creating an entity with nested arrays', () => {
   expect(exactUser.info).toHaveProperty('tags', [1, 2])
   expect(exactUser.info).toHaveProperty('documents', [])
 })
+
+test('model with properties that contain dots in them and can be considered a path', () => {
+  const db = factory({
+    user: {
+      'employee.id': primaryKey(datatype.uuid),
+    },
+  })
+
+  const dotUser = db.user.create({
+    'employee.id': 'abc-123',
+  })
+
+  expect(dotUser).toHaveProperty(['employee.id'], 'abc-123')
+})
--- a/test/model/create.test.ts
+++ b/test/model/create.test.ts
@@ -149,16 +149,16 @@ test('uses value getters when creating an entity with nested arrays', () => {
   expect(exactUser.info).toHaveProperty('documents', [])
 })
 
-test('model with properties that contain dots in them and can be considered a path', () => {
+test('supports property names with dots in model definition', () => {
   const db = factory({
     user: {
       'employee.id': primaryKey(datatype.uuid),
     },
   })
 
-  const dotUser = db.user.create({
+  const user = db.user.create({
     'employee.id': 'abc-123',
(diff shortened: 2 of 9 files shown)


src/relations/Relation.ts
```diff
@@ -66,7 +66,10 @@ export type ManyOf<ModelName extends KeyType> = Relation<
   any
 >
 
-export type RelationsMap = Record<string, Relation<any, any, any>>
+export type RelationsList = Array<{
+  path: string[]
```
kettanaito, line 70: What do you think if we name this property `propertyPath` to be aligned with how this is represented in the rest of the code?

test/model/create.test.ts
```diff
@@ -148,3 +148,17 @@ test('uses value getters when creating an entity with nested arrays', () => {
   expect(exactUser.info).toHaveProperty('tags', [1, 2])
   expect(exactUser.info).toHaveProperty('documents', [])
 })
+
+test('model with properties that contain dots in them and can be considered a path', () => {
```
kettanaito, line 152:
```suggestion
test('supports property names with dots in model definition', () => {
```

test/model/create.test.ts
```diff
@@ -148,3 +148,17 @@ test('uses value getters when creating an entity with nested arrays', () => {
...
+  const db = factory({
+    user: {
+      'employee.id': primaryKey(datatype.uuid),
+    },
+  })
+
+  const dotUser = db.user.create({
```
kettanaito, line 159:
```suggestion
  const user = db.user.create({
```

test/utils/definePropertyAtPath.test.ts
```diff
@@ -15,11 +15,31 @@ describe('definePropertyAtPath()', () => {
 
   it('defines a nested property at a given path', () => {
     const target: AnyObject = {}
-    definePropertyAtPath(target, 'a.b.c', {
+    definePropertyAtPath(target, ['a', 'b', 'c'], {
```
kettanaito, line 18: This is perhaps a silly question, but `propertyPath` here is not necessarily an array of strings, right?

If our intention is to use this paths for `lodash.get`/`lodash.set`, then it can also reference array indices in the path:

```js
definePropertyAtPath(target, ['a', 2, 'b'], value)
```

Numbers are also valid object keys, which makes me wonder if we should introduce a `PropertyPath` custom type that's a union of `string` and `number`:

```js
export type PropertyPath = Array<string | number>
```

What do you think about this, @VanTanev?

kettanaito (changes requested):
Thank you for this improvement, @VanTanev! 

The change looks fantastic. I've left a few comments that I'd appreciate your feedback on. I don't see any major blockers to get this done, just those comments. 

Looking forward to merging this.

src/relations/Relation.ts
VanTanev: I agree, done.

test/utils/definePropertyAtPath.test.ts
VanTanev: No, by ECMA standards an object key is ALWAYS a string. It will be cast to a string, regardless of what it was initially.

eg:
```ts
const obj = {}
const test = { 1: 'a', [obj]: 'b' }

Object.keys(test)
// ['1', '[object Object]']

test[1]
// a

test['1']
// a

test[obj]
// 'b'

test[{}]
// 'b'

test['[object Object]']
// 'b'
```

With that in mind, enforcing string keys in the type is good in my opinion, given that any value will ultimately be cast to a string. It's more straightforward and makes these implicit casts explicit.

VanTanev:
Thanks, @kettanaito, I made the suggested changes. Let me know if there is anything else, and thanks for the great lib!

test/utils/definePropertyAtPath.test.ts
VanTanev: Sorry, I just noticed I skipped a part of your question about array indices. 

It's true that `lodash.set`/`lodash.get` can set array indices, and numbers are valid there. However, my understanding is that `definePropertyAtPath` is meant to be used only for hashmap objects, and not for arrays, given the use of `Object.defineProperty()`.

kettanaito: At least not as of now. Using strict string type for object keys makes sense, thank you for explaining that.

kettanaito (approved):
Thank *you* for making this project great. It was a pleasure collaborating with you.

If you find any other tasks keen for your interest, feel free to raise a discussion/tackle them. I'd be happy to have you as a contributor.

VanTanev:
I don't want to rush you, but it would be super helpful to have a release cut with this PR mixed. Any guidance on when that might happen?

PR #155, opened 2021-10-30
fix(rest-handlers): parse number primary keys from query params

### LINKED ISSUE #154: Parse query params for number primary keys in REST handlers
```js
import { factory, primaryKey } from '@mswjs/data'

const db = factory({
  user: {
    id: primaryKey(Number), // Notice the primary key is of type Number
    firstName: String,
  },
})

db.user.toHandlers('rest')

db.user.create({
  id: 123,
  firstName: 'John',
})

// Will always return a 404 response
fetch('/users/123', {
  method: 'GET',
  headers: {
    'Content-Type': 'application/json',
  },
})
```

With the above example, when generating REST handlers, the following CRUD operations will not work as you would expect:

- `GET /users/:id` (where "id" is your model's primary key), returns a user by ID;
- `PUT /users/:id`, updates an existing user by ID;
- `DELETE /users/:id`, deletes an existing user by ID;

These CRUD operations don't work because the `where` clause is using the request `id` params as a String, where on the database we have it as a `Number`. These values don't match when querying the database and because of that, it will always return a 404 response.

To fix this, some additional logic should be added to the `generateRestHandlers` function in order to detect if the primary key is of type `Number` and then cast the request `id` param value accordingly when passing it to the `where` clause.

ruisaraiva19:
## GitHub

- Closes #154

## Changes

- Parse number primary keys from query params in REST handlers.

### DIFF
--- a/src/model/generateRestHandlers.ts
+++ b/src/model/generateRestHandlers.ts
@@ -7,7 +7,7 @@ import {
   PrimaryKeyType,
   ModelDefinition,
 } from '../glossary'
-import { GetQueryFor, QuerySelectorWhere } from '../query/queryTypes'
+import { QuerySelectorWhere, WeakQuerySelectorWhere } from '../query/queryTypes'
 import { OperationErrorType, OperationError } from '../errors/OperationError'
 import { findPrimaryKey } from '../utils/findPrimaryKey'
 import { PrimaryKey } from '../primaryKey'
@@ -23,10 +23,6 @@ class HTTPError extends OperationError<HTTPErrorType> {
     super(type, message)
     this.name = 'HTTPError'
   }
-}
-
-interface WeakQuerySelectorWhere<KeyType extends PrimaryKeyType> {
-  [key: string]: Partial<GetQueryFor<KeyType>>
 }
 
 type RequestParams<Key extends PrimaryKeyType> = {
@@ -119,7 +115,7 @@ export function parseQueryParams<ModelName extends string>(
 
 export function generateRestHandlers<
   Dictionary extends ModelDictionary,
-  ModelName extends string
+  ModelName extends string,
 >(
   modelName: ModelName,
   modelDefinition: ModelDefinition,
@@ -127,13 +123,17 @@ export function generateRestHandlers<
   baseUrl: string = '',
 ) {
   const primaryKey = findPrimaryKey(modelDefinition)!
-  const primaryKeyValue = (modelDefinition[primaryKey] as PrimaryKey).getValue() as PrimaryKeyType
+  const primaryKeyValue = (
+    modelDefinition[primaryKey] as PrimaryKey<PrimaryKeyType>
+  ).getValue()
   const modelPath = pluralize(modelName)
   const buildUrl = createUrlBuilder(baseUrl)
 
   function extractPrimaryKey(params: Record<string, string>): PrimaryKeyType {
-    const raw = params[primaryKey]
-    return isNaN(primaryKeyValue as any) ? raw : Number(raw)
+    const parameterValue = params[primaryKey]
+    return typeof primaryKeyValue === 'number'
+      ? Number(parameterValue)
+      : parameterValue
   }
 
   return [
@@ -166,7 +166,7 @@ export function generateRestHandlers<
           const id = extractPrimaryKey(req.params)
           const where: WeakQuerySelectorWhere<PrimaryKeyType> = {
             [primaryKey]: {
-              equals: id as any,
+              equals: id as string,
             },
           }
           const entity = model.findFirst({
@@ -192,7 +192,7 @@ export function generateRestHandlers<
           const id = extractPrimaryKey(req.params)
           const where: WeakQuerySelectorWhere<PrimaryKeyType> = {
             [primaryKey]: {
-              equals: id as any,
+              equals: id as string,
             },
           }
           const updatedEntity = model.update({
@@ -212,7 +212,7 @@ export function generateRestHandlers<
           const id = extractPrimaryKey(req.params)
           const where: WeakQuerySelectorWhere<PrimaryKeyType> = {
             [primaryKey]: {
-              equals: id as any,
+              equals: id as string,
             },
           }
           const deletedEntity = model.delete({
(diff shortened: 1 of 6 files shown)


test/model/toRestHandlers.test.ts
```diff
@@ -558,3 +562,170 @@ describe('DELETE /users/:id', () => {
     })
   })
 })
+
+describe('GET /todos/:id', () => {
+  it('handles a GET request to get a single entity', async () => {
```
kettanaito, line 567: We should put these tests into respective `describe` blocks that already exist. We may have to rename the model name to `user` to be consistent. What's important is to name the test case properly, so it's clear what use case we're asserting.

```js
it('handles a GET request to get an entity by a numeric primary key', ...)
```

src/model/generateRestHandlers.ts
```diff
@@ -126,6 +126,7 @@ export function generateRestHandlers<
   baseUrl: string = '',
 ) {
   const primaryKey = findPrimaryKey(modelDefinition)!
+  const primaryKeyValue = findPrimaryKeyValue(modelDefinition)!
```
kettanaito, line 129: We already know that `primaryKey` is the property name in `modelDefinition` that keeps the declaration of the primary key. We can access it directly instead of looking it up again in `findPrimaryKeyValue`:

```ts
const primaryKey = findPrimaryKey(modelDefinition)!
const primaryKeyValue = modelDefinition[primaryKey].getValue()
```

src/model/generateRestHandlers.ts
```diff
@@ -159,7 +160,7 @@ export function generateRestHandlers<
           const id = req.params[primaryKey]
           const where: WeakQuerySelectorWhere<typeof primaryKey> = {
             [primaryKey]: {
-              equals: id,
+              equals: typeof primaryKeyValue === 'number' ? +id : id as any,
```
kettanaito, line 163: What is the type error here so we cast the `id` value to `any`?

test/model/toRestHandlers.test.ts
ruisaraiva19: The `user` model name already exists. I added the `todos` model with a numeric primary key to add new assertions. That's the reason for this new `describe` block.

src/model/generateRestHandlers.ts
ruisaraiva19: > Type '{ [x: string]: { equals: string | number; }; }' is not assignable to type 'WeakQuerySelectorWhere<PrimaryKeyType>'.

CleanShot 2021-11-01 at 11 53 05

src/model/generateRestHandlers.ts
```diff
@@ -185,7 +186,7 @@ export function generateRestHandlers<
...
```
kettanaito, line 189: We seem to be repeating this logic quite a lot. Basically, whenever we're retrieving the actual primary key value from the `req.params`, we need to coerce it to number if that's its type in the model definition. 

Perhaps, let's abstract this coercion into a function?

```js
const primaryKeyValue = modelDefinition[primaryKey].getValue()

function extractPrimaryKey(params: Record<string, string>): PrimaryKeyType {
  const raw = params[primaryKey]
  return isNaN(primaryKeyValue) ? raw : Number(raw)
}

// ...
const id = extractPrimaryKey(req.params)
{
  [primaryKey]: { equals: id }
}
```

ruisaraiva19: That sounds good!

kettanaito (changes requested):
Hey, @ruisaraiva19. Thanks for this fix. I've left a few comments, would love to hear your opinion on those.

src/model/generateRestHandlers.ts
ruisaraiva19: This will show a TypeScript error related to type narrowing.
CleanShot 2021-11-01 at 11 58 51

ruisaraiva19: We can cast it's type like this:

```ts
const primaryKeyValue = (modelDefinition[primaryKey] as PrimaryKey).getValue()
```

src/model/generateRestHandlers.ts
kettanaito: It seems like `WeakQuerySelectorWhere<typeof primaryKey>` is an incorrect type, specifically, the generic is incorrect. `typeof primaryKey` will always be `string` because `primaryKey` holds the primary key property's _name_.

Instead, it should be `WeakQuerySelectorWhere<PrimaryKeyType>`, indicating that `id` is a union of all supported primary key types.

ruisaraiva19: It's the same thing. `typeof primaryKey` is `PrimaryKeyType`.

CleanShot 2021-11-01 at 12 24 07

src/model/generateRestHandlers.ts
ruisaraiva19: Changes pushed.

src/model/generateRestHandlers.ts
ruisaraiva19: Changes pushed.

kettanaito:
I'll try to put this on my list in the upcoming days. Once again thank you for your work on this.

kettanaito:
I've rebased the branch, removed the unnecessary usage of `faker` from the tests (we specify all the values anyway to assert them being returned from the handler). Also, split a single `toRestHandlers.test.ts` suite into 2 standalone test suites for readability. That'll help us manage any additional scenarios in the future.

I couldn't resolve the weak query type issue, TypeScript _is_ right, the union type will be converted to `StringQuery | NumberQuery` which seems not to know what the actual type of `id` will be. I've cast `id as string` to prevent any usage of `any`.

kettanaito (approved):
Thank you for fixing this, @ruisaraiva19! I'm happy to have you as a contributor! Keep up the great work, we need new team members 😉

test/model/toRestHandlers.test.ts
kettanaito: Got it. I've moved this entire scenario into its own test suite so it's easier to read it.

src/model/generateRestHandlers.ts
kettanaito: I've cast `id as string` so we at least don't resort to `any`.

ruisaraiva19:
@kettanaito I'm glad to provide value to the project!

PR #158, opened 2021-11-13
fix(types): relationships can be undefined

olivierwilkinson:
When a model is created any property that does not have an initial value uses the factory method to create one. Since the factories are typed to only allow certain primitive values it is not possible for those values to ever be `undefined`. For relational properties created through `oneOf` and `manyOf` this is not the case however, if they are not provided an initial value they are `undefined`.

For example in the following snippet:
```typescript
const db = factory({
  user: {
    id: primaryKey(String),
  },
  post: {
    id: primaryKey(String),
    author: oneOf('user'),
  },
})

const post = db.user.create();
```
`post.author` is `undefined` but it is not included in the property type.

This PR adds `undefined` to the possible value types for relationships.

### DIFF
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -199,13 +199,13 @@ export type Value<
     : // Extract value type from OneOf relations.
     Target[Key] extends OneOf<infer ModelName, infer Nullable>
     ? Nullable extends true
-      ? PublicEntity<Dictionary, ModelName> | null
-      : PublicEntity<Dictionary, ModelName>
+      ? PublicEntity<Dictionary, ModelName> | undefined | null
+      : PublicEntity<Dictionary, ModelName> | undefined
     : // Extract value type from ManyOf relations.
     Target[Key] extends ManyOf<infer ModelName, infer Nullable>
     ? Nullable extends true
-      ? PublicEntity<Dictionary, ModelName>[] | null
-      : PublicEntity<Dictionary, ModelName>[]
+      ? PublicEntity<Dictionary, ModelName>[] | undefined | null
+      : PublicEntity<Dictionary, ModelName>[] | undefined
     : // Account for primitive value getters because
     // native constructors (i.e. StringConstructor) satisfy
     // the "AnyObject" predicate below.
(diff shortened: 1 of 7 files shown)


src/glossary.ts
```diff
@@ -192,9 +192,9 @@ export type Value<
     ? ReturnType<Target[Key]['getValue']>
     : // Extract value type from relations.
     Target[Key] extends OneOf<infer ModelName>
-    ? PublicEntity<Dictionary, ModelName>
+    ? PublicEntity<Dictionary, ModelName> | undefined
     : Target[Key] extends ManyOf<infer ModelName>
-    ? PublicEntity<Dictionary, ModelName>[]
+    ? PublicEntity<Dictionary, ModelName>[] | undefined
```
kettanaito, line 197: Do we need to explicitly add `undefined` to this type? 

The initial values for the entities are always `Partial<Value<...>>`, which implies that its properties are only a subset of the model properties, making each property potentially undefined. Was there some use-case that made you add this change, @olivierwilkinson?

kettanaito: Is this change aimed exclusively at the proper annotation of the `post.author` value when inspecting it (as in, TypeScript says it's an existing relationship when it can also be undefined)?

kettanaito:
I've updated your feature branch against the latest main branch so it now includes your work on nullable properties.

src/glossary.ts
olivierwilkinson: Yes exactly it's for proper annotation of the `post.author` value. It cropped up when using entities as response data in msw graphql resolvers. For example:

```typescript
db.post.create() // post.author is undefined

type GetPost = {
  post: {
    id: string
    author: {
      id: string
    }
  }
}

server.use(
  graphql.query<GetPost, { id: string }>('GetPost', (req, res, ctx) => {
    const post = db.post.findFirst({
      where: { id: { equals: req.variables.id } },
      strict: true,
    })!

    // Typescript doesn't warn us that `post` doesn't match the GetPost type
    return res(ctx.data({ post })); 
  })
)
```

olivierwilkinson:
> I've updated your feature branch against the latest main branch so it now includes your work on nullable properties.

I'll fix the test errors that have been introduced now 👍

src/glossary.ts
kettanaito: Got it! Thanks.

kettanaito (approved):
Thank you for yet another awesome contribution, @olivierwilkinson!

PR #163, opened 2021-11-17
fix(executeQuery): support circular references when updating entities

### LINKED ISSUE #139: Cannot define model to have bidirectional relationships for child and parent
Using 0.7.0, I'm trying to create a hierarchy where Model A is a parent to Model B, which is in turn a parent to Model C. To preserve the links between them, a middle node like Model B would contain reciprocated references to its parent (A) and its child (C). Something, however, seems to be going awry:

https://codesandbox.io/s/dawn-sunset-4weoz?file=/src/App.js

Note that in the example above, commenting out either update will remove the error.

_Some context..._
Traversable, two-way relationships are a GraphQL implementation pattern. If you're exposing a schema, and you have a fragment coupled to a component, you might want that fragment to retrieve data up and/or down its hierarchy. It's possible to work around this limitation as the library stands (and this library's great), it's just messier.

olivierwilkinson:
closes #139 

When passing an object with circular references to `update` the `executeQuery` function fails: it passes the `query` object to `JSON.stringify` which throws a "cyclic object value" exception.

Circular references are created when a bi-directional relationship is made. For example:

```typescript
const db = factory({
  user: {
    id: primaryKey(String),
    posts: manyOf('post'),
  },
  post: {
    id: primaryKey(String),
    author: oneOf('user')
  },
});

const author = db.user.create({ id: 'user-1' })
const post = db.post.create({ id: 'post-1', author })
const nextAuthor = db.user.update({
  where: { id: { equals: author.id } },
  data: { posts: [post] },
})
```

If an entity that has been set up this way is passed to `update` it fails because of the above:
```typescript
db.post.update({
  where: { id: { equals: 'a-different-post' } },
  data: { author: nextAuthor },
});
```

As @tinleym talked about in #139, to mock some GraphQL responses these relationships need to be traversed in both directions. This makes these bi-directional setups required to properly mock those resolvers.

This PR adds tests for creating / using bi-directional relationships and fixes the error thrown by `executeQuery` by adding a `safeStringify` util that is used in place of `JSON.stringify`.

The `safeStringify` util replaces any circular references with a string representation instead. It's more or less a knock off of the [one from the MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value), but it also inserts some information about the entity it's replaced to keep the logs useful.

The stringified version of `nextAuthor` that would produced by `safeStringify` is:
```javascript
{ "id": "user-1", "posts": [{ "id": "post-1", "author": "Entity(type: user, id: user-1)" }] }
```

### DIFF
--- a/src/query/executeQuery.ts
+++ b/src/query/executeQuery.ts
@@ -11,6 +11,7 @@ import { paginateResults } from './paginateResults'
 import { Database } from '../db/Database'
 import { sortResults } from './sortResults'
 import { invariant } from 'outvariant'
+import { safeStringify } from '../utils/safeStringify'
 
 const log = debug('executeQuery')
 
@@ -50,7 +51,7 @@ export function executeQuery(
   query: WeakQuerySelector<any> & BulkQueryOptions<any>,
   db: Database<any>,
 ): Entity<any, any>[] {
-  log(`${JSON.stringify(query)} on "${modelName}"`)
+  log(`${safeStringify(query)} on "${modelName}"`)
   log('using primary key "%s"', primaryKey)
 
   const records = db.getModel(modelName)
@@ -74,7 +75,7 @@ export function executeQuery(
   const resultJson = Array.from(result.values())
 
   log(
-    `resolved query "${JSON.stringify(query)}" on "${modelName}" to`,
+    `resolved query "${safeStringify(query)}" on "${modelName}" to`,
     resultJson,
   )
 
--- a/src/utils/safeStringify.ts
+++ b/src/utils/safeStringify.ts
@@ -1,0 +1,22 @@
+import { ENTITY_TYPE, PRIMARY_KEY } from '../glossary'
+
+export function safeStringify(value: any) {
+  const seen = new WeakSet()
+
+  return JSON.stringify(value, (_, value) => {
+    if (typeof value !== 'object' || value === null) {
+      return value
+    }
+
+    if (seen.has(value)) {
+      const type = value[ENTITY_TYPE]
+      const primaryKey = value[PRIMARY_KEY]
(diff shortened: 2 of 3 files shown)


kettanaito:
Hey, @olivierwilkinson! Thank you for tackling this one.
I remember I was looking into this issue myself, need to dig up my old notes to see what I came up with. Your changes will certainly be useful.

olivierwilkinson:
> Hey, @olivierwilkinson! Thank you for tackling this one.
> 
> I remember I was looking into this issue myself, need to dig up my old notes to see what I came up with. Your changes will certainly be useful.

No worries 😁

I'd hit up against this one myself not too long ago so had an idea what the problem was. Very happy for you to go with a different solution if you had one in mind!👌

kettanaito:
Note to myself: keep notes referencable. Now to understand where exactly I've tackled this months ago...

kettanaito:
Thank you for working on this, @olivierwilkinson. Your changes are fantastic. 

I highly value your contributions! Could you please reach out to me [on Twitter](https://twitter.com/kettanaito)? I'd like to compensate you for your time working on the library and discuss if we could collaborate more. Thanks.

olivierwilkinson:
> Thank you for working on this, @olivierwilkinson. Your changes are fantastic. 
> 
> 
> 
> I highly value your contributions! Could you please reach out to me [on Twitter](https://twitter.com/kettanaito)? I'd like to compensate you for your time working on the library and discuss if we could collaborate more. Thanks.

@kettanaito thank you for your kind words! I would love to collaborate more and will send you a message now. My Twitter handle is olivierwilkins2 😁

Issue #89, opened 2021-04-28
Support nested where clauses


camchenry:
It would be nice to support nested conjunctions/disjunctions (AND/OR) for creating complex where clause queries.

Example from the [Prisma docs](https://www.prisma.io/docs/concepts/components/prisma-client/crud#filter-by-multiple-field-values):
```ts
const users = await prisma.user.findMany({
  where: {
    OR: [
      {
        name: {
          startsWith: 'E',
        },
      },
      {
        AND: {
          profileViews: {
            gt: 0,
          },
          role: {
            equals: 'ADMIN',
          },
        },
      },
    ],
  },
})
```

Haven't looked into implementing this, but would probably require checking for a special property like `AND` or `OR` and handling the property accordingly.

camchenry:
If it's easier, the library could support non-recursive queries first, and add recursive functionality later? So these queries would be supported:

```ts
// Supported currently
const users = db.user.findMany({
  where: {
    name: {
      equals: "something",
    },
  },
});

// NEW: single layer disjunction (OR)
const users = db.user.findMany({
  where: {
    OR: [
      {
        name: {
          equals: "something",
        },
      },
      {
        email: {
          contains: "@example.com",
        },
      },
    ],
  },
});

// NEW: single layer conjunction (AND)
const users = db.user.findMany({
  where: {
    AND: [
      {
        name: {
          equals: "something",
        },
      },
      {
        email: {
          contains: "@example.com",
        },
      },
    ],
  },
});
```

kettanaito:
Hey, @camchenry. That's a good proposal. I believe your recursive `AND` example is missing an array, right?

```diff
const users = await prisma.user.findMany({
  where: {
    OR: [
      {
        name: {
          startsWith: 'E',
        },
      },
      {
-        AND: {
+        AND: [{
          profileViews: {
            gt: 0,
          },
          role: {
            equals: 'ADMIN',
          },
        },
-      },
+      }],
    ],
  },
})
```

If I understand that API correctly, both `OR` and `AND` support a list of values but treat matches in those list differently:

- OR is basically `list.some()`
- AND is `list.every()`

kettanaito:
Please, would you be interested in lending us a hand once more?

I think the barebone for this feature would be to extend the `compileQuery` function somewhere along these lines:

https://github.com/mswjs/data/blob/154d5b556970d9a7c09ba9c49e86ec7887620b07/src/query/compileQuery.ts#L28

[...]

PR #169, opened 2021-11-25
fix(create): allow the field with type array to be an array of objects

### LINKED ISSUE #165: Fields with type `Array` cannot be an array of object
I have example code here:
https://github.com/nvh95/msw-example/blob/d144e5c98f165c15b8758e1157fa478952307d7f/src/mocks/browser.js#L4-L29
Given I have this schema:
```js
const db = factory({
  user: {
    id: primaryKey(Number),
    email: String,
    additional_data: Array,
  },
});
```
Basically, I have a field `additional_data` which is an `Array`. If I create array of primitive type like `string` or `number`, it works fine. 
```js
db.user.create({
  ...
  additional_data: [1, 2, 3 ],
});
```
image

But if `additional_data` is an array of object, `@mswjs/data` seems cannot process it and returns an empty array
```js
db.user.create({
  ...
  additional_data: [
    {
      key: "hobby",
      value: "code",
    },
    {
      key: "phone",
      value: "12345",
    },
  ],
});
```
Screenshot 2021-11-19 at 15 50 10

I have code to reproduce this issue in https://github.com/nvh95/msw-example/blob/msw-example-2021.11.19
```bash
git clone git@github.com:nvh95/msw-example.git
git checkout msw-example-2021.11.19
npm i
npm start
```

@kettanaito Can you take a look? Is this an issue or I misuse `additional_data: Array` (I refer to #113). (definitely I don't want to use `manyOf` here). Thank you very much.

roertbb:
- Fixes #165

I stumbled upon an issue mentioned in https://github.com/mswjs/data/issues/165 when using an array of objects for one of the fields of the model. After downgrading the version to v0.7.2 it worked fine. 

If it's a regression as mentioned in https://github.com/mswjs/data/issues/165#issuecomment-974645341, removing the additional check for primitive values in case of an array should suffice. However, I'm not sure if it could affect the feature of nullable properties from https://github.com/mswjs/data/pull/143.

I just started using the library and I'm not fully aware of some parts of it as well as the internals 😉  Let me know if I'm missing something. With some guidance, I believe I could fix that one 😄 

Btw. Huge thanks for maintaining `msw` and making the process of creating tests so nice!

### DIFF
--- a/test/model/create.test.ts
+++ b/test/model/create.test.ts
@@ -13,7 +13,7 @@ test('creates a new entity', () => {
   // Without any arguments a new entity is seeded
   // using the value getters defined in the model.
   const randomUser = db.user.create()
-  expect(randomUser).toHaveProperty('id', userId)
+  expect(randomUser.id).toEqual(userId)
 })
 
 test('creates a new entity with initial values', () => {
@@ -27,23 +27,23 @@ test('creates a new entity with initial values', () => {
   const exactUser = db.user.create({
     id: 'abc-123',
   })
-  expect(exactUser).toHaveProperty('id', 'abc-123')
+  expect(exactUser.id).toEqual('abc-123')
 })
 
 test('creates a new entity with an array property', () => {
   const db = factory({
     user: {
       id: primaryKey(datatype.uuid),
-      arrayProp: Array,
+      favouriteNumbers: Array,
     },
   })
 
   const exactUser = db.user.create({
     id: 'abc-123',
-    arrayProp: [1, 2, 3],
+    favouriteNumbers: [1, 2, 3],
   })
-  expect(exactUser).toHaveProperty('id', 'abc-123')
-  expect(exactUser).toHaveProperty('arrayProp', [1, 2, 3])
+  expect(exactUser.id).toEqual('abc-123')
+  expect(exactUser.favouriteNumbers).toEqual([1, 2, 3])
 })
 
 test('creates a new entity with an array property with array of objects assigned', () => {
@@ -70,8 +70,8 @@ test('creates a new entity with an array property with array of objects assigned
     notes: exampleNotes,
   })
 
-  expect(exactUser).toHaveProperty('id', 'abc-123')
-  expect(exactUser).toHaveProperty('notes', exampleNotes)
+  expect(exactUser.id).toEqual('abc-123')
+  expect(exactUser.notes).toEqual(exampleNotes)
 })
 
 test('creates a new entity with nullable properties', () => {
@@ -92,9 +92,9 @@ test('creates a new entity with nullable properties', () => {
     name: null,
   })
 
-  expect(user).toHaveProperty('name', null)
-  expect(user).toHaveProperty('age', null)
-  expect(user.address).toHaveProperty('number', null)
+  expect(user.name).toEqual(null)
+  expect(user.age).toEqual(null)
+  expect(user.address.number).toEqual(null)
 })
 
 test('supports nested objects in the model definition', () => {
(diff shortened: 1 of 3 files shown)


test/model/create.test.ts
```diff
@@ -46,6 +46,34 @@ test('creates a new entity with an array property', () => {
...
+    },
+  ]
+
+  const exactUser = db.user.create({
+    id: 'abc-123',
+    arrayProp: arrayOfObjects,
+  })
+
+  expect(exactUser).toHaveProperty('id', 'abc-123')
+  expect(exactUser).toHaveProperty('arrayProp', arrayOfObjects)
```
kettanaito, line 74: I recommend accessing the property directly on the target object instead of using `.toHaveProperty` method. One of the benefits of doing so is getting type validation when accessing a non-existing property (i.e. a typo, or when the property got removed). `.toHaveProperty` accepts a general `string` as the first argument. That string may not exist as a property on the target object. 

```suggestion
  expect(exactUser.arrayProp).toEqual(arrayOfObjects)
```

test/model/create.test.ts
```diff
@@ -46,6 +46,34 @@ test('creates a new entity with an array property', () => {
   expect(exactUser).toHaveProperty('arrayProp', [1, 2, 3])
 })
 
+test('creates a new entity with an array property with array of objects assigned', () => {
+  const db = factory({
+    user: {
+      id: primaryKey(datatype.uuid),
+      arrayProp: Array,
```
kettanaito, line 53: Let's come up with a property name that's closer to how people would model real-world data with arrays. This would make our test shareable as usage examples, which is a great feat for a test. 

```suggestion
      notes: Array,
```

Perhaps, let's consider `user.notes`, where each note is, basically, a list of `key/value` pairs (as you have it now)?

kettanaito (changes requested):
Hey, @roertbb. Thank you for preparing this fix! The job you did here is simply fantastic, I think that's the right approach to solve this issue. 

I've left a few comments, mainly to make this change even better. Let me know what you think about those. I can't wait to get this bug fix to the next release. Thank you!

test/model/create.test.ts
roertbb: Make sense! I've adjusted the example as suggested in https://github.com/mswjs/data/pull/169/commits/a29fa1c69d10a6870612601aa0d7320e63e97f63

test/model/create.test.ts
roertbb: Valid point as well! 

Initially, I wanted to make the test consistent with the convention that was already there, that's why I went with `toHaveProperty`. After adjusting the matchers for that case I've decided to apply it for other tests too (https://github.com/mswjs/data/pull/169/commits/b192a5bae0960eaf9557bb0a82732235555e22e8) - let me know if you have some concerns about that. 

I've noticed an issue with inferring the structure of the nested resource, that's why I went with https://github.com/mswjs/data/pull/169/commits/5cd820da591e8ceb794e6bf6438e00b6d994daa6#diff-9e87669b89f42eb50b386f8e76828b8c646e42c60598b3d5cd50d9b36152ba2bR179

test/model/create.test.ts
kettanaito: Fantastic! Thank you for tackling this.

test/model/create.test.ts
kettanaito: My apologies, the test suites are not yet aligned on this pattern. Yet another pull request idea to carry out 😉

> I've noticed an issue with inferring the structure of the nested resource
I think I've fixed that in #171 but that may be an entirely different issue.

kettanaito (approved):
@roertbb, thank you for making this library better! Welcome to contributors 🎉 

If you find any other task interesting, please don't hesitate to sparkle a discussion. I know we could use your expertise on some of the issues. What's great is that a lot of them need to be discussed/brainstormed to find the optimal API to achieve what's requested (i.e. #89). I'd love to hear your opinion on those. 

Thank you!

VanTanev:
Thanks!

PR #171, opened 2021-11-26
fix(types): annotate nested relationships

kettanaito:
## Current behavior

```ts
const db = factory({
  user: {
    id: primaryKey(String),
    country: oneOf('country'),
    stats: {
      revision: oneOf('revision'),
    },
  },
  country: {
    code: primaryKey(String),
  },
  revision: {
    id: primaryKey(String),
    updatedAt: Number,
  },
})

const user = db.user.create()
user.country.code.toUpperCase()

user.stats.revision?.id // TypeError
user.stats.revision?.updatedAt.toFixed() // TypeError
```

A nested relationship is not annotated because the `Value` generic for the nested properties passed the incorrect `Dictionary` generic:

https://github.com/mswjs/data/blob/0970e7ed5a9094d6618ed6563f8bb3a00d5c9ecc/src/glossary.ts#L216

This causes the `user.stats.revision` to be type of `Value<undefined, OneOf<...>>`. That `undefined` means that the relationship value cannot be inferred when using `PublicEntity` later on:

https://github.com/mswjs/data/blob/0970e7ed5a9094d6618ed6563f8bb3a00d5c9ecc/src/glossary.ts#L203

https://github.com/mswjs/data/blob/0970e7ed5a9094d6618ed6563f8bb3a00d5c9ecc/src/glossary.ts#L68

`Dictionary[ModelName]` yields `undefined`.

### DIFF
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -213,7 +213,7 @@ export type Value<
     ? ReturnType<Target[Key]>
     : // Handle nested objects.
     Target[Key] extends AnyObject
-    ? Partial<Value<Target[Key], Target>>
+    ? Partial<Value<Target[Key], Dictionary>>
     : // Otherwise, return the return type of primitive value getters.
       ReturnType<Target[Key]>
 }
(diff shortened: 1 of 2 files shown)


kettanaito:
This may be a regression of #113.

PR #177, opened 2021-12-05
Move debug to dependencies

Methuselah96:
`debug` is imported in the built package so it should be a dependency instead of a devDependency.


kettanaito (approved):
Thank you for improving this, @Methuselah96! Welcome to contributors 🎉

PR #180, opened 2021-12-15
fix(types): nullable relations are never undefined

olivierwilkinson:
I made a mistake when updating #158 after we merged #143, it is not actually possible for nullable relationships to be `undefined`, they are `null` instead. Sorry I let that slip by!

This is purely for proper annotation similarly to #158 

It is not possible for the value of nullable relations to be undefined because:
- if the relation does not have an initial value it defaults to null.
- it is not possible to update the relation's value to be undefined after the fact.


kettanaito (approved):
Thank you for spotting this, @olivierwilkinson! A great fix.

PR #187, opened 2022-01-26
test(findMany): add failing test for querying by null

kettanaito:
- Fixes #186

### DIFF
--- a/test/model/findMany.test.ts
+++ b/test/model/findMany.test.ts
@@ -1,5 +1,5 @@
 import { datatype } from 'faker'
-import { factory, primaryKey } from '../../src'
+import { factory, primaryKey, nullable } from '../../src'
 import { OperationErrorType } from '../../src/errors/OperationError'
 import { getThrownError } from '../testUtils'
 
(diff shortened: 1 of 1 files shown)


test/model/findMany.test.ts
```diff
@@ -76,3 +76,26 @@ test('returns an empty array when not found matching entities', () => {
...
+  })
+
+  const john = db.user.create({ id: 'john' })
+  db.user.create({ id: 'katy', organizationId: 'org-1' })
+  const clark = db.user.create({ id: 'clark' })
+
+  const users = db.user.findMany({
+    where: {
+      organizationId: {
+        equals: null,
```
kettanaito, line 95: Querying nullable fields by `null` is a legal operation.

kettanaito:
## Insights

I suspect that the type produced by `Value` is not correct:

https://github.com/mswjs/data/blob/d32f15461fc0ef915da6709de62e9e9156b8a234/src/glossary.ts#L211-L212

First of all, it misses the `| null` union type. But even when added, it still remains just a plain model value type without the `null` union. I have no idea why this happens, neither I have the tools to debug this with the current types complexity of Data. 

Eventually, a `null` check must be introduced to the `queryTypes.ts` to produce a proper query type:

https://github.com/mswjs/data/blob/d32f15461fc0ef915da6709de62e9e9156b8a234/src/query/queryTypes.ts#L88-L89

kettanaito:
## Update

1.0 most likely fixes this. The task is now to see if this test exists in the new test suites and if it doesn't, add it.

kettanaito:
## Released: v1.0.1 🎉

This has been released in v1.0.1.

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v1.0.1)
- 📦 [View on npm](https://www.npmjs.com/package/@msw/data/v/1.0.1)

```
npm i @msw/data@latest
```

---

PR #201, opened 2022-04-06
feat(manyOf): return empty array for non-nullable without value

### LINKED ISSUE #199: Feature Request: manyOf can optionally return empty array instead of undefined
## Use Case
GraphQL schemas may require an array type, in which case returning undefined will throw an error. The current workaround is to explicitly extract all fields from an Entity and append an or fallback to all manyOf fields. It would be nice if this wasn't the case.

## API Change
Change the manyOf(entityName: string) declaration to manyOf(entityName: string, emptyArrayFallback?: boolean)

Happy to submit a PR if this fits the goals of the library!

kettanaito:
- Fixes #199 

## Changes

- Non-nullable `manyOf` relation now returns an empty array `[]` when it has no value (previously: `undefined`).
- Other behaviors are intact. Nullable relations _always_ return `null` when they have no value.

### DIFF
--- a/src/model/defineRelationalProperties.ts
+++ b/src/model/defineRelationalProperties.ts
@@ -9,7 +9,7 @@ import {
   PRIMARY_KEY,
   Value,
 } from '../glossary'
-import { Relation, RelationsList } from '../relations/Relation'
+import { RelationKind, RelationsList } from '../relations/Relation'
 
 const log = debug('defineRelationalProperties')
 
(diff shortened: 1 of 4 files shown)


src/model/defineRelationalProperties.ts
```diff
@@ -49,28 +49,36 @@ export function defineRelationalProperties(
...
+      log('has no references but is nullable, applying a getter to null...')
       relation.resolveWith(entity, null)
+      continue
+    }
+
+    if (relation.kind === RelationKind.ManyOf) {
+      log(
+        'has no references but is a non-nullable "manyOf" relationship, applying a getter to []...',
+      )
+      relation.resolveWith(entity, [])
```
kettanaito, line 78: The actual change.

borodean:
Shouldn't the typing be also updated like this?

```diff
diff --git a/src/glossary.ts b/src/glossary.ts
index 925cc4a..63326ab 100644
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -230,7 +230,7 @@ export type Value<
     Target[Key] extends ManyOf<infer ModelName, infer Nullable>
     ? Nullable extends true
       ? PublicEntity<Dictionary, ModelName>[] | null
-      : PublicEntity<Dictionary, ModelName>[] | undefined
+      : PublicEntity<Dictionary, ModelName>[]
     : // Account for primitive value getters because
     // native constructors (i.e. StringConstructor) satisfy
     // the "AnyObject" predicate below.
```

kettanaito:
Hi, @borodean. I believe you're right, I've missed the type definition updates. Thank you for spotting this!

PR #209, opened 2022-05-11
fix(manyOf): exclude undefined from non-nullable without value

borodean:
This PR fixes the typings to match the behavioural update in https://github.com/mswjs/data/pull/201 (#199).

### DIFF
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -230,7 +230,7 @@ export type Value<
     Target[Key] extends ManyOf<infer ModelName, infer Nullable>
     ? Nullable extends true
       ? PublicEntity<Dictionary, ModelName>[] | null
-      : PublicEntity<Dictionary, ModelName>[] | undefined
+      : PublicEntity<Dictionary, ModelName>[]
     : // Account for primitive value getters because
     // native constructors (i.e. StringConstructor) satisfy
     // the "AnyObject" predicate below.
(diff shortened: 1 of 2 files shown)


kettanaito:
The CI is failing on some typings tests. I think we should also update those tests to reflect this change:

https://github.com/mswjs/data/blob/9f55935f572207598ea1ad00b35f577f0ee8f170/test/model/relationalProperties.test-d.ts#L23-L24

The return type of a non-nullable `manyOf` is no longer potentially undefined. Let's remove this line.

kettanaito (approved):
A fantastic addition, @borodean! 🎉 Thank you.

borodean:
> Welcome to the contributors, @borodean! 🥳
> Let me know if any other tasks in Data interest you. I know I could use your help.

🤗

I'm just figuring out MSW for myself these days, but I'd be happy to lend my hand to an open source project. I followed you on Twitter, we can talk there.

kettanaito:
Looking forward to that! 

I don't expect any commitment from contributors but instead, I want them to find areas they are curious about and help them improve there through my open source. We've got Service Workers, Node.js implementation of requests, data modeling, cookie handling, and so much more. 

I'm currently overwhelmed by all the tasks around the projects, so it's always great to have another person to collaborate with. Feel free to DM me if you find some time to get to improving MSW!

Issue #217, opened 2022-05-20
Update Faker Dependency and Imports


shawntax:
In January 2022, there was an incident with the original mainter of Faker and a new [community-maintained project](https://github.com/faker-js/faker) has been created. More info on [that here](https://fakerjs.dev/about/announcements/2022-01-14.html).

I've already [updated references to Faker](https://github.com/mswjs/data/pull/216) in the README, but since this library has `faker` as a dependency for internal testing, it should be updated along with any references as described [here.](https://fakerjs.dev/migration-guide-v5/#remove-all-references-to-faker-from-your-project-the-new-package-is-located-at-faker-js-faker=)

PR #218, opened 2022-05-21
Update to Faker v6

shawntax:
For some background on what happened to the Faker package, see [this announcement](https://fakerjs.dev/about/announcements/2022-01-14.html). TL;DR: the previous maintainer of the Faker went rouge and deleted the package. A new group of mainters stepped up to bring back the project and keep it going, moving it to a new GitHub org. 

As such, this PR does a few things:

1) Removes the deleted `faker` dev dependency along with `@types/faker`. It then adds the `v6` of the community-maintained package under `@faker-js/faker`. v6 ships [with TS support](https://fakerjs.dev/guide/#typescript-support), so there's no need for a `@types` dependency.

2) Updates import statements and usage of the `Faker` package in the test files as described [in the migration guide](https://fakerjs.dev/guide/#migrating-from-faker-v5-to-v6). Side note here: even though the tests look like they currently only depend on the english locale, I am not updating the import statements because Faker is only being used in a test environment (basing that on [this answer](https://github.com/faker-js/faker/discussions/887#discussioncomment-2655425) from one of the new package maintainers).

3) Updates the code examples in the README to reflect the recommened way to do imports in v6.

Fixes https://github.com/mswjs/data/issues/217


kettanaito (approved):
A truly marvelous pull request to review. Thank you once again, @shawntax!

Issue #203, opened 2022-04-08
Support nested nullable objects


kettanaito:
- Originated from #202 

This has to be supported:

```js
const db = factory({
  user: {
    id: primaryKey(String),
    permissions: nullable(() => ({
      id: String,
    })),
  },
})
```

But it's not. Recursive nullable objects as well. No primary key needed.

Aloysb:
@kettanaito I'll give this a crack 😉

Aloysb:
See PR #219 👍 😃

yishayweb:
Hi guys,

I opened a new PR for solving this issue with an api that doesn't use a nullable getter as @kettanaito asked
Can someone please take a look at it [here](https://github.com/mswjs/data/pull/257) ?

@kettanaito @abitbetterthanyesterday please let me know if this solution is acceptable and if so I will add some notes about it in the documentation

Thanks

kettanaito:
## Released: v1.0.0 🎉

This has been released in v1.0.0.

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v1.0.0)
- 📦 [View on npm](https://www.npmjs.com/package/@msw/data/v/1.0.0)

Get these changes by running the following command:

```
npm i @msw/data@latest
```

---

_Predictable release automation by [Release](https://github.com/ossjs/release)_.

PR #219, opened 2022-05-22
Feat/#203 support nullable nested objects

aloysb:
This is regarding issue #203 

I have extended the existing `ModelValueType` and the function `isModelValueType`  to allow for `object` type.

I also added a new test case covering the scenario described in #203 .

It was indeed failing at the beginning, the object would be null no matter whether or not you passed to create an existing object. 
This was caused by `isModelValueType` that would return `false` if the object was not a Primitive or an Array of Primitives.

The test is now passing, and I've also tested it manually.

Note that this is my PR for this project and unfortunately the tests were not all green out of the box. So I manually checked where these functions/types were used to ensure I didn't break existing functionalities.

Please don't hesitate to let me know if anything needs to be changed.

### DIFF
--- a/src/utils/isModelValueType.ts
+++ b/src/utils/isModelValueType.ts
@@ -10,5 +10,9 @@ function isPrimitiveValueType(value: any): value is PrimitiveValueType {
 }
 
 export function isModelValueType(value: any): value is ModelValueType {
-  return isPrimitiveValueType(value) || Array.isArray(value)
+  return (
+    isPrimitiveValueType(value) ||
+    Array.isArray(value) ||
+    typeof value === 'object'
+  )
 }
--- a/README.md
+++ b/README.md
@@ -472,10 +472,16 @@ When using Typescript, you can manually set the type of the property when it can
 ```typescript
 import { factory, primaryKey, nullable } from '@mswjs/data'
 
+type Car = {
+  brand: string
+  model: string
+}
+
 const db = factory({
   user: {
     id: primaryKey(String),
     age: nullable<number>(() => null),
+    car: nullable<Car>(() => null),
   },
 })
 ```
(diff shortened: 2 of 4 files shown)


src/utils/isModelValueType.ts
```diff
@@ -10,5 +10,9 @@ function isPrimitiveValueType(value: any): value is PrimitiveValueType {
 }
 
 export function isModelValueType(value: any): value is ModelValueType {
-  return isPrimitiveValueType(value) || Array.isArray(value)
+  return (
+    isPrimitiveValueType(value) ||
+    Array.isArray(value) ||
+    typeof value === 'object'
```
Aloysb, line 16: Unsure if this is too permissive?

JasonShin (approved):

JasonShin:
Nice one, can we get it merged?

roertbb:
Hi there! 👋 I've also tried to tackle that one. I think that there may be some parts missing, when for example we'd like to have some default value for nullable object property - for example 

```ts
const db = factory({
  user: {
    id: primaryKey(faker.datatype.uuid),
    address: nullable(() => ({
      street: () => "Wall Street",
      number: nullable<number>(() => null),
    })),
  },
});

const user = db.user.create();
```

we may want to have something like

```json
{
    id: "...",
    address: {
        street: "Wall Street",
        number: null
    }
}
```
for `user`.

I created https://github.com/mswjs/data/pull/231/files with some investigation I've made some time ago and finished today. Maybe you could take a look there, because you're familiar with the problem and we could figure out some working solution together?

Aloysb:
Hi @roertbb, sorry I was quite busy, I'll have a look at your code and get back to you ASAP :+1:

yishay-at-bay:
Hi,
Any progress on this one?
It would really upgrade this already amazing package :)
Thanks !!

PR #231, opened 2022-07-03
Nullable object properties

### LINKED ISSUE #203: Support nested nullable objects

roertbb:
Hi there! 👋 I stumbled upon that case 2 or 3 times and decide to give it a shot to solve my problem and most likely help some other great folks as well 😄 

I've read the hints from https://github.com/mswjs/data/discussions/202, extended `ModelValueType`, adjusted the generation of nested object values within nullable nested objects and also adjusted types here and there to properly handle nullable nested objects.

Sharing some context behind technical decisions/changes I've made:
- I decided to add `isGetterFunctionReturningObject` to `NullableProperty` - it indicates if we should generate properties for nested nullable objects in `src/model/parseModelDefinition.ts`. It's needed when we have some getter function returning object with nested properties in factory and we're not providing any value for these nested object properties during model creation (case is covered in `nullable object property > when object getter is provided in factory definition > defaults to value provided in factory when not set during model creation` test)
- I also created `getDefinition` function to "unwrap" the getter value from `NullableProperty` when we want to get the default values of nested properties within nullable object properties
- I've extended the type for `Value` in `glossary.ts` to properly infer types in case of some object is passed as generic type. It was also needed to make `nulllable<Date>(() => null)` to work properly (you can find some examples in `test/query/date.test.ts`)
- changes in `src/query/queryTypes.ts` were required when querying dates with `lte` / `gte` (case is covered in `ignores entities with missing values when querying using date` test)

I tried to provide API without a getter function (as mentioned in https://github.com/mswjs/data/issues/203), however I failed on providing proper typing for it (some raw PoC can be found in https://github.com/roertbb/data/pull/1). Recently, decided to neglect that idea and try out extending existing `NullableProperty` (which is using getter function) to incorporate objects there.

Working on that one was a pretty good journey into depths of `mswjs/data`, I think I understood quite some of them, but I still may be missing some other parts - I'd be really grateful for guidance on what I could've missed or how I could make it more elegant and readable 😉

I also wrote couple of tests that served me as a test cases. These are not extensively testing all the possible use-cases, but should be a good foundation. If you have other test-cases on your mind, if you feel that it will be more readable to structure tests some other way - please share some feedback 😉 

Resolves #203

### DIFF
--- a/src/glossary.ts
+++ b/src/glossary.ts
@@ -19,7 +19,10 @@ export type KeyType = string | number | symbol
 export type AnyObject = Record<KeyType, any>
 export type PrimaryKeyType = string | number
 export type PrimitiveValueType = string | number | boolean | Date
-export type ModelValueType = PrimitiveValueType | PrimitiveValueType[]
+export type ModelValueType =
+  | PrimitiveValueType
+  | PrimitiveValueType[]
+  | AnyObject
 export type ModelValueTypeGetter = () => ModelValueType
 
 export type ModelDefinition = Record<string, ModelDefinitionValue>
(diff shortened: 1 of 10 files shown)


kettanaito:
## Released: v1.0.0 🎉

This has been released in v1.0.0.

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v1.0.0)
- 📦 [View on npm](https://www.npmjs.com/package/@msw/data/v/1.0.0)

```
npm i @msw/data@latest
```

---

PR #237, opened 2022-09-02
feat(createMany): add createMany api

themagickoala:
There have been a few times where I've found it would be useful to pass an array of model definitions to create, so I've added a createMany api. Happy to take direction on improvements, love what you're doing with this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,7 @@ drop(db)
 Each model has the following methods:
 
 - [`create()`](#create)
+- [`createMany()](#createMany)
 - [`findFirst()`](#findFirst)
 - [`findMany()`](#findMany)
 - [`count()`](#count)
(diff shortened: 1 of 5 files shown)


shairez:
@kettanaito thanks for this tool!
btw is there a way we can merge this PR? 
I can help if there's a need to update the code in some way
Thanks!

kettanaito:
Thanks, @shairez. Much appreciated. This has been a ton of work!

themagickoala:
Thanks Artem, can't wait to try this out!

kettanaito:
Thanks! Let me know how it goes. See the [`.createMany()`](https://github.com/mswjs/data?tab=readme-ov-file#createmanycount-initialvaluesfactory) API for more details.

PR #240, opened 2022-09-20
feat: add basic OR query

louisch:
Hi, I wanted to take a crack at this issue here: https://github.com/mswjs/data/issues/89

I've made a very basic POC that just adds the ability to do an OR query on the first level, haven't made it work recursively yet, only made sure that the type checks pass and the test I created passes.

Just wanted to make sure that @kettanaito was okay with the general direction of this PR, and if you have any improvement or suggestions you have for me, especially about the types, because quite frankly I'm not a Typescript expert, and the code style as well.

Next would probably be making this actually recursive, and then trying to add AND as well.


nathanpower:
Any chance of this getting merged?

PR #257, opened 2022-12-29
feat: created nullable object option without getter

yishayweb:
This PR is for supporting nullable objects without a getter function (mentioned in [issue 203](https://github.com/mswjs/data/issues/203))

I took some inspiration from @roertbb solution, but solved it a little bit differently. I haven't changed the ModelValueType definition but rather made a new NullableObject class that accepts a value of the type NestedModelDefinition, since this is in fact what it should receive. It also accepts a second arg that indicates if the NullableObject should default to null (I made it so that the complete object definition will always be given while creating an instance of the class)

In general, I think that @kettanaito request [here](https://github.com/mswjs/data/issues/203#issue-1197709742) to implement it without a getter function makes a lot of sense as it is more consistent with regular nested objects (because we don't define them with a getter function as well)

### DIFF
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -149,7 +149,6 @@ function createModelApi<
     update({ strict, ...query }) {
       const results = executeQuery(modelName, primaryKey, query, db)
       const prevRecord = first(results)
-      // console.log('results: ', results)
 
       if (!prevRecord) {
         if (strict) {
--- a/src/factory.ts
+++ b/src/factory.ts
@@ -149,6 +149,7 @@ function createModelApi<
     update({ strict, ...query }) {
       const results = executeQuery(modelName, primaryKey, query, db)
       const prevRecord = first(results)
+      // console.log('results: ', results)
 
       if (!prevRecord) {
         if (strict) {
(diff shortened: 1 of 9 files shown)


jchris (approved):
thanks this looks great. I am new to the project so take my approval with a grain of salt

yishayweb:
Thanks @jchris !!
I would also like to get an approval from someone that is more familiar with the code, but so far no luck with that :(
@kettanaito I saw you mentioned that you are rewriting the package to release a new major version. But is there a chance to insert this change to the current major version? Thanks !!

diogomiguel:
This seems dead so @kettanaito Any chance you can review this as requested by OP? Quite keen on this functionality. Thanks 🙏

kettanaito (approved):
Thanks for your work on this, @yishayweb 👏 Welcome to contributors.

kettanaito:
## Released: v0.14.0 🎉

This has been released in v0.14.0!

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v0.14.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/data/v/0.14.0)

---

PR #263, opened 2023-02-02
Allow passing `Object` constructor to a field definition - useful for JSON columns

StanleySathler:


src/utils/isModelValueType.ts
```diff
@@ -1,14 +1,16 @@
+import isObjectLike from 'lodash/isObjectLike'
 import { ModelValueType, PrimitiveValueType } from '../glossary'
 
 function isPrimitiveValueType(value: any): value is PrimitiveValueType {
   return (
     typeof value === 'string' ||
     typeof value === 'number' ||
     typeof value === 'boolean' ||
+    isObjectLike(value) ||
```
kettanaito, line 9: Can we use an already existing `isObject` utility?

https://github.com/mswjs/data/blob/5be9ca36f560bc2977ae68ea96651b4cd7ad0250/src/utils/isObject.ts

test/model/create.test.ts
```diff
@@ -30,6 +30,22 @@ test('creates a new entity with initial values', () => {
   expect(exactUser.id).toEqual('abc-123')
 })
 
+test('creates a new entity with an object property', () => {
```
kettanaito, line 33: Let's add another test with the same setup but the `.create()` call where `.settings` wasn't specified. What would be the expected output in that case?

kettanaito (changes requested):
Hey, @StanleySathler! Thank you so much for proposing this. I really appreciate your feedback and ideas on the library, it's simply fantastic. 

I left a few comments around this change, mainly about reusing existing utilities. Please let me know what you think about that.

MarcLopezAvila:
Hi! I think this feature has been asked for a while and some people are looking forward to it, if this pr makes it work it doesn't seem like a very big change. @kettanaito is it possible to inlcude this in a further release (before the all rewrite I think it is being done)? We are block by this feature other people and would be so great to have it supported. Thanks!

StanleySathler:
Hey @MarcLopezAvila 👋 

I believe all changes Artem suggested should be pretty straightforward, which I'll take some time today to address. Then @kettanaito can decide the best way to release this. 🚀

MarcLopezAvila:
Sounds great, thank you so much!

test/model/create.test.ts
StanleySathler: Good Q. If we set `settings` as non-nullable, but don't specify a value to `.create()` method, I guess it should throw?

src/utils/isModelValueType.ts
StanleySathler: Done - [704c03a](https://github.com/mswjs/data/pull/263/commits/704c03aec83a92026ba707cbb7350ea4f1afc7c2)

test/model/create.test.ts
StanleySathler: Done, lemme know if that's not what you suggested - [02584dc](https://github.com/mswjs/data/pull/263/commits/02584dc0b4a02f108d97aa2bcabe0a54c2c4c68d)

StanleySathler:
@kettanaito your feedback should be addressed now 🚀 Can you recheck?

src/utils/isModelValueType.ts
```diff
@@ -1,10 +1,13 @@
+import isObjectLike from 'lodash/isObjectLike'
```
MarcLopezAvila, line 1: Unused import here

MarcLopezAvila (changes requested):

src/utils/isModelValueType.ts
StanleySathler: Nice catch - turns out we had a test set as `.only` too, which is also fixed now 🚀

MarcLopezAvila (approved):

MarcLopezAvila:
Hi @kettanaito, could you check if we can merge this change pls? Thank you very much!

Eadroma:
can we have an update about this please ?

kalilovsky:
Hello, we are using msw-data, but we really need this feature, can you please update us ?

src/utils/isModelValueType.ts
```diff
@@ -1,10 +1,12 @@
 import { ModelValueType, PrimitiveValueType } from '../glossary'
+import { isObject } from './isObject'
 
 function isPrimitiveValueType(value: any): value is PrimitiveValueType {
   return (
     typeof value === 'string' ||
     typeof value === 'number' ||
     typeof value === 'boolean' ||
+    isObject(value) ||
```
kettanaito, line 9: How will this work with nested objects? Those will be treated as primitive value types with this change, preventing the library from recursively analyzing their fields.

kettanaito: Our test suite will confirm that. Will run it to verify...

kettanaito (approved):
Thank you for working on this, @StanleySathler 👏

kettanaito:
## Released: v0.16.2 🎉

This has been released in v0.16.2!

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v0.16.2)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/data/v/0.16.2)

---

Issue #267, opened 2023-03-02
Using wildcard/parameters in baseUrl in toHandlers method


timkolotov:
I recently started using MSW for both jest and storybook tests.
In our project, we use relative paths for the API, so the resulting URL depends on the environment the call is made. E.g. if storybook is run on `http://127.0.0.1:6006/` this host will be used in the API call. But it can also be run on `http://localhost:6006/`

So in MSW handlers, I used baseUrl with parameters such as `http://:hostname/:lang/api/v1/` and it works perfectly. Now I wanted to reduce the amount of boilerplate and add `mswjs/data` with automatic generation of handles, but it seems to be not possible anymore as the URL constructor in `createUrlBuilder` fails to parse such a string as that is not a valid URL.

Is there any other way for manipulating the resulting paths in handlers, like just adding prefixes? Or the only way is to manually add all handlers?

kettanaito:
Hey, @timkolotov. I think this should be fixed on the library's side. The `createUrlBuilder` function, as you've rightfully mentioned, should support path strings as well.

https://github.com/mswjs/data/blob/5be9ca36f560bc2977ae68ea96651b4cd7ad0250/src/model/generateRestHandlers.ts#L32-L37

From the look of it, I don't see any implications of supporting that. Would you be interested in opening a pull request with this feature?

timkolotov:
Yeah, will do. Probably this weekend.

timkolotov:
@kettanaito hey, I created a PR for this. It seems I can't request a review for it, so just FYI :slightly_smiling_face:

kettanaito:
Thanks, @timkolotov! Not sure what's with the review request permissions but I will look into it regardless.

PR #268, opened 2023-03-05
feat: support relative paths when generating rest handlers

timkolotov:
Removes usage of URL constructor in `createUrlBuilder` which is used for generating REST handlers.
Approach with strings concatenation allows us to use relative paths without specific hostnames, provide prefixes, and also use parameters in the baseUrl that will be handled by MSW.

I also added removing trailing slashes from the baseUrl, as previously it could be passed with and without it and won't affect the result, so I keep it like it was to avoid breaking changes.
Additionally updated tests for the function: removed the leading slash. Because `buildUrl` calls inside `generateRestHandlers` have no leading shashes so tests weren't exactly correct.

(ref: #267)

### DIFF
--- a/src/model/generateRestHandlers.ts
+++ b/src/model/generateRestHandlers.ts
@@ -31,8 +31,9 @@ type RequestParams<Key extends PrimaryKeyType> = {
 
 export function createUrlBuilder(baseUrl?: string) {
   return (path: string) => {
-    const url = new URL(path, baseUrl || 'http://localhost')
-    return baseUrl ? url.toString() : url.pathname
+    // For the previous implementation trailing slash didn't matter, we must keep it this way for backward compatibility
+    const normalizedBaseUrl = baseUrl && baseUrl.slice(-1) === '/' ? baseUrl.slice(0, -1) : baseUrl || '';
+    return `${normalizedBaseUrl}/${path}`
   }
 }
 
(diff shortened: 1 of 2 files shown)


timkolotov:
@kettanaito Something seems to be very wrong with the CI. I don't know much about CI in GitHub, but perhaps `actions/checkout` must be updated?
I had such an error locally in another project, was related to OpenSSL and Node...

kettanaito:
Yeah, the issue seems to be not related to your changes. I will look into that.

timkolotov:
@kettanaito anything here?

timkolotov:
@kettanaito fixed it, I guess https://github.com/mswjs/data/pull/270 (still can't request the review)

kettanaito:
Thanks for fixing that, @timkolotov. Could you please rebase this branch? The CI issue should be gone then.

timkolotov:
Seems all good now :smiley:

kettanaito (approved):
🚀

kettanaito:
## Released: v0.12.0 🎉

This has been released in v0.12.0!

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v0.12.0)
- 📦 [npm package](https://www.npmjs.com/package/@mswjs/data/v/0.12.0)

---

PR #312, opened 2025-05-29
docs: fix faker uuid usage

robwalkerco:
Faker provides uuid values from `faker.string.uuid` - Docs reference - https://fakerjs.dev/api/string.html#uuid


kettanaito (approved):
Thanks!

PR #323, opened 2025-10-02
fix: implement `RelationError`, improve unique relations validation

### LINKED ISSUE #307: Describe approach of error handling
Currently, the documentation doesn't state any approach of how to handle errors.

Topics to cover:
1. What is the format of errors? Do they have different types (extensions of `Error` object)? Do they have unique error codes, like [Prisma does](https://www.prisma.io/docs/orm/reference/error-reference#error-codes)?
2. Does a model have built-in validators depending on provided meta-data (like type)? Can users define custom validators? Can they customize errors? If so, can they customize only error messages, or they can return a custom structure? Or should they validate data passed to the model beforehand (using Zod, for example), because the library's scope will be intentionally small?
3. Is there a difference of handling validation errors (via a dedicated `validate` method, for example) and other errors? Or are they all [handled the same way, like in Prisma](https://www.prisma.io/docs/orm/prisma-client/debugging-and-troubleshooting/handling-exceptions-and-errors)?

Even though the library will be rewritten, as mentioned in #285, this information will still be necessary.

kettanaito:
- Closes #307
- Documents error handling in `README.md`.
- Implements a `RelationError` class.
- Refactors the `OperationError` to include `code` instead of specific operation names.
- Validates unique relations for the lack of other owners referencing the same foreign record(s) (for one-way unique relations).
- Adds missing tests for unique relations.

## Todos

- [x] Use `code` for `OperationError` instead of `.operationName`. Make it an enum. 
- [x] Change the tests to assert the right errors are being thrown. 
- [x] Add unique many-to relation tests.
  - one-to-many unique relations tests are present. The many-to relations cannot be unique by design; they implied that many records may reference many other records.

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,9 +432,8 @@ Data provides multiple different error classes to help you differentiate and han
 
 ### `OperationError`
 
-- `operationName` `<string>`, the name of the errored operation (e.g. "create", "updateMany", etc.);
-- `info` `<object>`, additional operation information (often the operation arguments);
-- `cause` `<Error>`, a reference to the original thrown error.
+- `code` `<OperationErrorCode>`, the error code describing the failed operation;
+- `cause` `<Error>` (_optional_), a reference to the original thrown error.
 
 Thrown whenever performing a record operation fails. For example:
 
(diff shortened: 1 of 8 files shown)


kettanaito:
## Released: v1.0.1 🎉

This has been released in v1.0.1.

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v1.0.1)
- 📦 [View on npm](https://www.npmjs.com/package/@msw/data/v/1.0.1)

```
npm i @msw/data@latest
```

---

PR #341, opened 2025-10-14
Add failing test case for adding items to one-to-many relations in `update()`

Turbo87:
It seems that the `update()` is not actually being applied if the updated field is a relation instead of just a regular `z.array()`.

I might be doing something wrong, but I can't get this use case to work... 😅 

## Todos

- [ ] Fix the [failing test](https://github.com/mswjs/data/pull/341#issuecomment-3403478915).
- [ ] Uncomment the other new tests for updates through the owner and have them passing.

### DIFF
--- a/tests/relations/one-to-many.test.ts
+++ b/tests/relations/one-to-many.test.ts
@@ -165,7 +165,7 @@ it('supports updating foreign relations through the owner updates', async () => 
     })
 })
 
-it.skip('supports updating one-to-many relations by pushing new records to the relational array', async () => {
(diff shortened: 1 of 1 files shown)


rewweRrr:
Is there a workaround?

> @Turbo87, reassigning the entire relational property instead of pushing a new value should work as-is right now to unblock your migration.
> 
> ```ts
> data(post) {
>   post.comments = [...post.comments, newComment]
> }
> ```
> 
> Could you give this a try?

PR #350, opened 2025-11-29
fix: many relation updates via memoized arrays and guarded hooks

### LINKED ISSUE #346: v1.1.2 Not able to use update() to add entry into many() relation
In the following example, the final post only has one (1) comment, whereas I would expect it to have two (2).

```ts
const postSchema = z.object({
  get comments() {
    return z.array(commentSchema)
  },
})
const commentSchema = z.object({
  text: z.string(),
})

const posts = new Collection({ schema: postSchema })
const comments = new Collection({ schema: commentSchema })

posts.defineRelations(({ many }) => ({
  comments: many(comments),
}))

const post = await posts.create({ 
  comments: [await comments.create({ text: 'First!' })]
});

const comment2 = await comments.create({ text: 'Second!' });

const updatedPost = await posts.update(post, {
  data(p) {
    p.comments = [  ...post.comments, comment2 ];
    // p.comments.push(comment2) also doesn't work
  }})

console.log(updatedPost.comments.length)
```

I'm using Bun 1.3.2, msw 2.12.2, msw/data 1.1.2 and zod 4.1.12.

unadlib:
Fix #346 

Fix many() relation updates so array assignments and element-level mutations keep foreignKeys and inverse links in sync, guarding the “through relation” early hook from array indices.
  - Maintain a stable resolved array for many relations so push/splice produce patches and relational updates get observed.
  - Add regression tests covering many-relation array reassignment and push to ensure new foreign records appear and order is preserved.

## Solution

This fix uses memoization: the many() relation now maintains a stable resolved array (cached and mutated in place) so Mutative can draft it and emit patches on push/splice. We didn’t materialize getters into plain arrays before drafting.

  - Why memoize the resolved array? Mutative only emits patches for mutations on draftable references. The relation getter was returning a fresh array each time, so draft.comments was not the same object Mutative proxied, making push invisible (no patches). By caching a stable array per owner record and reusing it, Mutative can wrap that array, see push/splice, and generate patches. We still update that array’s contents on resolve so it reflects current foreign records while keeping the reference stable.
  - Why not materialize getters up-front? Pre-flattening all relational getters into plain arrays before calling createDraft would lose the getter semantics during the draft, risk breaking circular relations, and require extra restoration logic. Memoization keeps getters intact and keeps the minimal surface change (just stabilizes the returned reference).
  - Early hook guard change: The “through relation” early hook was firing for array index paths (e.g., comments.0), treating pushes as nested foreign updates and short-circuiting owner updates. Adding the path.length > relationPath.length and “next segment not a number” guard ensures only genuine nested foreign updates are intercepted; array-level updates now flow to the relation handlers.
  - Foreign key sync on array updates: We now explicitly recompute/add/remove foreignKeys (and inverse links) on relation-level array reassignment and on index-level patches, so both spread (draft.comments = [...]) and push stay consistent with inverse relations and uniqueness checks.

### DIFF
--- a/src/relation.ts
+++ b/src/relation.ts
@@ -196,8 +196,10 @@ export abstract class Relation {
       const update = event.data
 
       if (
+        update.path.length > path.length &&
         path.every((key, index) => key === update.path[index]) &&
-        !isRecord(update.nextValue)
+        !isRecord(update.nextValue) &&
+        typeof update.path[path.length] !== 'number'
       ) {
         event.preventDefault()
         event.stopImmediatePropagation()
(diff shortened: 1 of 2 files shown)


kettanaito:
## Released: v1.1.3 🎉

This has been released in v1.1.3.

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v1.1.3)
- 📦 [View on npm](https://www.npmjs.com/package/@msw/data/v/1.1.3)

```
npm i @msw/data@latest
```

---

PR #358, opened 2026-04-19
fix: support relational updates via array push

### LINKED ISSUE #346: v1.1.2 Not able to use update() to add entry into many() relation

kettanaito:
- Fixes #346
- Closes #350

(push/splice): mutative can't see through getters (mutative#157).
  - collection.ts — new #buildDraftInput: clones via cloneWithInternals (type-preserving), then overwrites each relation path with the live resolved value (keeps foreign records'
  kPrimaryKey). Passed to createDraft instead of prevRecord.
  - collection.ts — patch loop collapses per-index patches (['posts', N] or ['posts', 'length']) into a single {path: ['posts'], nextValue: <final array>} event → reuses Fix A's Many
  handler. Deeper paths (['posts', 0, 'title']) pass through.
  - relation.ts — early hook extended for Many: when next segment is numeric, resolve the indexed foreign record and propagate the trailing path to just that record (was broadcasting to
   all).

### DIFF
--- a/src/collection.ts
+++ b/src/collection.ts
@@ -761,8 +761,17 @@ export class Collection<Schema extends StandardSchemaV1> {
       prevRecord,
     )
 
+    /**
+     * @note Build a draft input where relational getters are replaced with
+     * plain data descriptors holding their resolved values. `mutative` cannot
+     * observe mutations through getters (https://github.com/unadlib/mutative/issues/157),
+     * so a `push` on `draft.posts` is lost unless `posts` is a plain property.
+     * Live foreign record references (with their internal symbols) are preserved.
+     */
+    const draftInput = this.#buildDraftInput(prevRecord)
+
     const [maybeNextRecord, patches, inversePatches] = await createDraft(
-      prevRecord,
+      draftInput,
       updateData,
       {
         strict: false,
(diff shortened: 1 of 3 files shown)


kettanaito:
## Released: v1.1.3 🎉

This has been released in v1.1.3.

- 📄 [**Release notes**](https://github.com/mswjs/data/releases/tag/v1.1.3)
- 📦 [View on npm](https://www.npmjs.com/package/@msw/data/v/1.1.3)

```
npm i @msw/data@latest
```

---

PR #364, opened 2026-05-13
fix: support self-referencing relations

jrjohnson:
Failing tests for #363 - when a model references itself updating one side updates the other in unexpected ways.

- Fixes #363


tests/relations/one-to-one.test.ts
```diff
@@ -651,3 +651,37 @@ it('removes relation listeners when the owner record is deleted', async () => {
     'Detaches relation listeners when the owner record is deleted',
   ).toBe(baseline)
 })
+
+it('updates a self referencing one to one relation', async () => {
```
kettanaito: The test name seems to be misleading: there's no `.update()` happening anywhere. From what I can see, these are declaration tests. 


```suggestion
it('creates a self referencing one-to-one relation', async () => {
```

Maybe?

